PR #4, opened 2020-02-04
Tyler Porter

tylerpporter:
Submission for Snow Day Challenge

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,3 @@
+class House
+
+end 
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -15,5 +15,9 @@ class House
 
   def above_market_range?
     @price > 500000 ? true : false
-  end 
+  end
+
+  def rooms_from_category(category)
+    @rooms.select {|room| room.category == category}
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -8,4 +8,8 @@ class House
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
+    @rooms << room
+  end 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -11,5 +11,9 @@ class House
 
   def add_room(room)
     @rooms << room
+  end
+
+  def above_market_range?
+    @price > 500000 ? true : false
   end 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,10 +1,11 @@
 class House
-  attr_reader :price, :address
+  attr_reader :price, :address, :rooms
 
   def initialize(price, address)
     # I'm going to defend this now... why would we ever need to store price data for a house as anything other than an integer?
-    # I don't think it's necessary to preserve anything other than numbers in this attribute. 
+    # I don't think it's necessary to preserve anything other than numbers in this attribute.
     @price = price.gsub(/[^0-9]/, "").to_i
     @address = address
+    @rooms = []
   end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,5 +1,5 @@
 class House
-  attr_reader :price, :address, :rooms
+  attr_reader :price, :address, :rooms, :details 
 
   def initialize(price, address)
     # I'm going to defend this now... why would we ever need to store price data for a house as anything other than an integer?
@@ -8,6 +8,7 @@ class House
     @price = price.gsub(/[^0-9]/, "").to_i
     @address = address
     @rooms = []
+    @details = {"price"=>@price, "address"=>@address}
   end
 
   def add_room(room)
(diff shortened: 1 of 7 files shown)


lib/house.rb
```diff
@@ -34,4 +34,12 @@ def price_per_square_foot
   def rooms_sorted_by_area
     @rooms.sort_by {|room| room.area}
   end
+
```
tylerpporter, line 36: I know there is a much more efficient way to do this but I was drawing a blank at the time so this is the route I went.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
+class House
+  attr_reader :price, :address, :rooms, :details
+
+  def initialize(price, address)
+    # I don't think it's necessary to preserve anything other than fixnums in this attribute.
+    # Regexp superior to just deleting the $ to account for any other invalid character
+    @price = price.gsub(/[^0-9]/, "").to_i
```
timomitchel, line 7: I like this! Thinking about scale here. Nice work

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+    @address = address
+    @rooms = []
+    @details = {"price"=>@price, "address"=>@address}
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_range?
```
timomitchel, line 17: Remember to name methods exactly what they are in the spec. Otherwise, they won't pass our internal tests we use to grade these.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+    @price > 500000 
+  end
+
+  def rooms_from_category(category)
+    @rooms.select {|room| room.category == category}
```
timomitchel, line 22: Great job choosing the right enumerable here.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def area
+    @rooms.sum {|room| room.area}
```
timomitchel, line 26: Nice use of `sum` with the block to get the job done.

lib/house.rb
timomitchel: This works. It's the first one I reached to as well when doing this method. However, you can use `sort` here with a block and it will avoid having to call reverse.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+  def rooms_sorted_by_area
+    @rooms.sort_by {|room| room.area}.reverse
+  end
+
+  def rooms_by_category
+    keys = @rooms.map(&:category).uniq
+    values = keys.map do |key|
+      @rooms.select {|room| room.category == key}
+    end
+    keys.zip(values).to_h
```
timomitchel, line 42: This is an interesting approach! You're doing a little bit more work than you need to though. Check out the `group_by` method/enum.

test/house_test.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  def test_it_can_add_rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_it_knows_it_its_above_market_range
```
timomitchel, line 28: Nice job making this test robust and testing multiple scenarios.

test/house_test.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    assert_equal false, house_2.above_market_range?
+
+    house_3 = House.new("$500001", "234 cat st")
+
+    assert house_3.above_market_range?
+  end
+
+  def test_it_can_organize_rooms_by_category
+    @house.add_room(@room_1)
```
timomitchel, line 42: You could think about adding this pattern to your setup method in order to DRY this code up.

test/house_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal [@room_1, @room_2], @house.rooms_from_category(:bedroom)
```
timomitchel, line 47: Nice job testing all of the categories.

test/house_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    assert_equal [@room_4, @room_3, @room_2, @room_1], @house.rooms_sorted_by_area
+  end
+
+  def test_it_can_return_a_hash_of_rooms_by_category
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    rooms_by_category = {:bedroom=>[@room_1, @room_2], :living_room=>[@room_3], :basement=>[@room_4]}
```
timomitchel, line 91: The convention when using a variable in a test in this way is to use `expected` as the name of the variable, but this is also acceptable!

timomitchel (commented):
Great job on this challenge Tyler! You're looking strong. Make sure to help out your classmates that aren't as comfortable yet. Nothing solidifies our mastery like teaching and putting code into words!

PR #5, opened 2020-02-04
Anthony Gomes

tony-gomes:
Completed Iteration 1
Completed Iteration 2
Completed Iteration 3
Completed Iteration 4

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -34,10 +34,9 @@ class House
     (@price / area.to_f).round(2)
   end
 
-  # def rooms_sorted_by_area
-  #   require 'pry'; binding.pry
-  #   @rooms.each { |room| room.area }.sort { |low, high| high <=> low }
-  # end
+  def rooms_sorted_by_area
+    @rooms.sort_by { |room| room.area }.reverse
+  end
 
   def rooms_by_category
     rooms_hash = {}
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -20,7 +20,7 @@ class House
   end
 
   def area
-    @rooms.map { |room| room.length * room.width }.sum
+    @rooms.map { |room| room.area }.sum
   end
 
   def details
@@ -28,5 +28,26 @@ class House
     details['price'] = @price
     details['address'] = @address
     details
+  end
+
+  def price_per_square_foot
+    (@price / area.to_f).round(2)
+  end
+
+  # def rooms_sorted_by_area
+  #   require 'pry'; binding.pry
+  #   @rooms.each { |room| room.area }.sort { |low, high| high <=> low }
+  # end
+
+  def rooms_by_category
+    rooms_hash = {}
+    @rooms.each do |room|
+      if !rooms_hash.key?(room.category)
+        rooms_hash[room.category] = [room]
+      else
+        rooms_hash[room.category] << room
+      end
+    end
+    rooms_hash
   end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -10,4 +10,23 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    @price > 500_000
+  end
+
+  def rooms_from_category(category)
+    @rooms.select { |room| room.category == category }
+  end
+
+  def area
+    @rooms.map { |room| room.length * room.width }.sum
+  end
+
+  def details
+    details = {}
+    details['price'] = @price
+    details['address'] = @address
+    details
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,13 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price.tr('^0-9', '').to_i
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,5 +1,5 @@
 class Room
-  attr_reader :category
+  attr_reader :category, :length, :width
 
   def initialize(category, length, width)
     @category = category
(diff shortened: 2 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,52 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price.tr('^0-9', '').to_i
```
aziobrow, line 5: this is good! there are quite a few ways you could do this without regex too. 🙂

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    @price > 500_000
```
aziobrow, line 15: beautiful!

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.select { |room| room.category == category }
```
aziobrow, line 19: great enumerable for this job!

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def area
+    @rooms.map { |room| room.area }.sum
```
aziobrow, line 23: remember, sum is an enumerable too. 

```
def area
    @room.sum { |room| room.area }
end
```

The way you have it, you're technically enumerating twice (one for map and one for sum), which is unnecessary.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def details
+    details = {}
+    details['price'] = @price
+    details['address'] = @address
+    details
```
aziobrow, line 30: very nice!

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def price_per_square_foot
+    (@price / area.to_f).round(2)
```
aziobrow, line 34: beautiful. nice job using the return value from another method here.

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by { |room| room.area }.reverse
```
aziobrow, line 38: perfect!

lib/house.rb
```diff
@@ -0,0 +1,52 @@
...
+    rooms_hash = {}
+    @rooms.each do |room|
+      if !rooms_hash.key?(room.category)
+        rooms_hash[room.category] = [room]
+      else
+        rooms_hash[room.category] << room
+      end
+    end
+    rooms_hash
+  end
```
aziobrow, line 51: This is really great, Tony! 

You could also look into `group_by`, for the future... http://gregpark.io/blog/ruby-group-by/
, which would change this into a one-liner.

aziobrow (commented):
Tony! So proud of you-- a perfect 4, you didn't miss a thing.  Great, fantastic, wonderful work!

Left some inline comments with specifics/suggestions. 

lib/room.rb
```diff
@@ -1,12 +1,22 @@
-class Rom
+class Room
+  attr_reader :category, :length, :width
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
-    @width = width
+    @length = length.to_i
```
aziobrow, line 6: not a big deal, but length was already an integer and didn't need to be converted.

lib/room.rb
```diff
@@ -1,12 +1,22 @@
...
+    @width = width.to_i
+    @status = false
```
aziobrow, line 8: we'd probably call this something like `@is_painted`, since `@status` could be the status of a great many things...

test/house_test.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def test_it_starts_no_rooms_and_can_add_rooms
+    assert_equal [], @house.rooms
+    @house.add_room(@room1)
+    @house.add_room(@room2)
+    assert_equal [@room1, @room2], @house.rooms
+  end
+
+  def test_it_can_determine_market_average
+    refute @house.above_market_average?
```
aziobrow, line 29: remember that if you have multiple possible outcomes, you should test them all. i'd expect to see another house that was above market average, and the relevant assertion for that.

test/house_test.rb
```diff
@@ -0,0 +1,88 @@
...
+    @house.add_room(@room3)
+    @house.add_room(@room4)
+    rooms_hash = {
+      bedroom: [@room1, @room2],
+      living_room: [@room3],
+      basement: [@room4]
+    }
+
+    assert_equal rooms_hash, @house.rooms_by_category
+  end
```
aziobrow, line 87: great work

test/room_test.rb
```diff
@@ -3,22 +3,32 @@
...
+
+  def test_it_starts_painted_and_can_be_painted
+    refute @room1.is_painted?
+    refute @room2.is_painted?
+
+    @room1.paint
+    @room2.paint
+
+    assert @room1.is_painted?
+    assert @room2.is_painted?
```
aziobrow, line 32: 👍 

Since these rooms are fundamentally the same, you probably only need to paint one room to feel pretty confident that that method works.

test/room_test.rb
```diff
@@ -3,22 +3,32 @@
 require './lib/room'
 
 class RoomTest < Minitest::Test
+  def setup
+    @room1 = Room.new(:bedroom, 10, '13')
+    @room2 = Room.new(:living_room, 15, '12')
```
aziobrow, line 8: since the tests were already created without a setup, it probably cost you more time to move these rooms into a setup than just to add them into the one test that you had to write... it's not a big deal, but do think about only making setup methods when they make sense.

PR #6, opened 2020-02-04
Andrew Reid

reid-andrew:
Iterations 1-4 Complete on Snow Day Challenge

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,7 @@
+class House
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -7,4 +7,8 @@ class House
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
+    @rooms << room
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -33,7 +33,7 @@ class House
   end
 
   def rooms_sorted_by_area
-    @rooms.sort_by {|room| room.area}
+    @rooms.sort_by {|room| room.area}.reverse
   end
 
   def rooms_by_category
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -15,4 +15,10 @@ class House
   def above_market_average?
     @price > 500000 ? true : false
   end
+
+  def rooms_from_category(category)
+    category_rooms = @rooms.select do |room|
+      room.category == category
+    end
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,11 +1,15 @@
 class House
 
-  attr_reader :price, :address, :rooms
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,50 @@
+class House
+
+  attr_reader :price, :address, :rooms, :details
+
+  def initialize(price, address)
+    @price = price.gsub("$", "").to_i
+    @address = address
+    @rooms = []
+    @details = {
```
memcmahon, line 9: Since we are already storing `@price` and `@address`, we don't need to store the details on the object, this could be a standalone method to cut down on memory used.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+      "address" => @address
+    }
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    @price > 500000 ? true : false
```
memcmahon, line 20: Great use of a ternary statement!

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.select {|room| room.category == category}
```
memcmahon, line 24: good use of `select`

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def price_per_square_foot
+    (@price.to_f / area).round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by {|room| room.area}.reverse
+  end
+
+  def rooms_by_category
```
memcmahon, line 39: You're logic here looks great and great job using the rooms_from_category method!

test/house_test.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  def test_rooms_can_be_added
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_if_it_is_above_market_average
+    assert_equal false, @house.above_market_average?
```
memcmahon, line 39: If a method returns either true *or* false, it is good to have a test that confirms either outcome

test/house_test.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  def test_it_has_rooms_from_category
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+    assert_equal [@room_1, @room_2], @house.rooms_from_category(:bedroom)
```
memcmahon, line 47: This is a great test - looking at multiple scenarios, including what should be returned if none are found.

memcmahon (commented):
Hey Andy - this looks great!  Take a look at my notes, but overall, it looks like you are on track for where we would expect you to be 🙂

PR #7, opened 2020-02-04
Maxwell Baird

Maxwell-Baird:


lib/house.rb
```diff
@@ -0,0 +1,62 @@
+
+class House
+
+  attr_reader :address, :price, :rooms, :details, :rooms_by_category
+
+  def initialize(price_parameter, address_parameter)
+    @price = price_parameter.delete_prefix('$').to_i
+    @address = address_parameter
+    @rooms = []
+    @details = {"price" => @price, "address" => @address}
```
memcmahon, line 10: Since we are already storing `@price` and `@address`, we don't need to store the details on the object, this could be a standalone method to cut down on memory used.

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    @rooms_by_category = {}
```
memcmahon, line 11: Similarly, we don't need to store rooms by category, because a room stores its own category. This would be better as a separate method.

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    @rooms << room
+    if @rooms_by_category.key?(room.category)
+      @rooms_by_category[room.category] << room
+    else
+      @rooms_by_category[room.category] = [room]
+    end
+  end
+
+  def above_market_average?
+    if @price > 500000
```
memcmahon, line 24: The logic here is perfect - you could also rely on the return value of the condition instead of building out a full if statement:

```ruby
def above_market_average?
  @price > 500000
end
```

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category_parameter)
```
memcmahon, line 31: Great use of `each` - you could also use `find_all`

```ruby
@rooms.find_all do |room|
  room.category == category
end
```

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    array_of_rooms_from_category = []
+    @rooms.each do |room|
+      if room.category == category_parameter
+        array_of_rooms_from_category << room
+      end
+    end
+    array_of_rooms_from_category
+  end
+
+  def area
```
memcmahon, line 41: Again, great use of `each` and you could also use sum:

```ruby
@rooms.sum do |room|
  room.area
end
```

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    area_sum = 0
+    @rooms.each do |room|
+      area_sum = area_sum + room.area
+    end
+    area_sum
+  end
+
+  def price_per_square_foot
+    price_per_square = 0.0
```
memcmahon, line 50: You don't _need_ this line :)

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    price_per_square = @price.to_f / area.to_f
+    price_per_square.round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by do |room|
+      -room.area
```
memcmahon, line 57: Great use of math to change the way this sorts :)

lib/room.rb
```diff
@@ -1,12 +1,24 @@
-class Rom
+class Room
+
+  attr_reader :length, :width, :category
```
memcmahon, line 3: Since we never need to call `length` or `width` as methods, they do not need to be attr_readers

test/house_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    house.add_room(room_1)
+    house.add_room(room_2)
+    assert_equal [room_1,room_2], house.rooms
+  end
+
+  def test_if_house_is_above_market_average?
+    house1 = House.new("$400000", "123 sugar lane")
+    assert_equal false, house1.above_market_average?
+    house1 = House.new("$600000", "111 fake lane")
+    assert_equal true, house1.above_market_average?
```
memcmahon, line 37: Great job testing both possible outcomes of this method!

memcmahon (commented):
Hey Max - this looks great!  Take a look at my notes, but overall, it looks like you are on track for where we would expect you to be 🙂

PR #9, opened 2020-02-04
David Tran

DavidTTran:
Created additional tests and methods for the room class. Also added house class and accompanying tests, that return 'price', 'address', 'rooms', 'area'

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -5,6 +5,7 @@ class House
     @price = price[1..-1].to_i
     @address = address
     @rooms = []
+    @total_area = 0
     @details = {"price" => @price,
                 "address" => @address
                }
@@ -19,20 +20,32 @@ class House
   end
 
   def rooms_from_category(category)
-    sorted_by_category = []
+    rooms_in_category = []
     @rooms.each do |room|
       if room.category == category
-        sorted_by_category << room
+        rooms_in_category << room
       end
     end
-    sorted_by_category
+    rooms_in_category
   end
 
   def area
-    total_area = 0
     @rooms.each do |room|
-      total_area += room.area
+      @total_area += room.area
     end
-    total_area
+    @total_area
+  end
+
+  def price_per_square_foot
+    area
+    (@price / @total_area.to_f).round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by(&:area)
+  end
+
+  def rooms_by_category
+    @rooms.sort_by {|room| [:category, room.category]}
   end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,9 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price[1..-1].to_i
+    @address = address
+    @rooms = []
+  end
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -42,7 +42,7 @@ class House
   end
 
   def rooms_sorted_by_area
-    @rooms.sort_by(&:area)
+    @rooms.sort_by(&:area).reverse!
   end
 
   def rooms_by_category
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -6,4 +6,8 @@ class House
     @address = address
     @rooms = []
   end
+
+  def add_room(room_object)
+    @rooms << room_object
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -30,15 +30,11 @@ class House
   end
 
   def area
-    @rooms.each do |room|
-      @total_area += room.area
-    end
-    @total_area
+    @rooms.sum {|room| room.area}
   end
 
   def price_per_square_foot
-    area
-    (@price / @total_area.to_f).round(2)
+    (@price / area.to_f).round(2)
   end
 
   def rooms_sorted_by_area
@@ -46,6 +42,8 @@ class House
   end
 
   def rooms_by_category
-    @rooms.sort_by {|room| [:category, room.category]}
+    @rooms.sort_by {|key, value|}.each do |room|
+      room.category
+    end
   end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,13 +1,38 @@
 class House
-  attr_reader :price, :address, :rooms
+  attr_reader :price, :address, :rooms, :details
 
   def initialize(price, address)
     @price = price[1..-1].to_i
     @address = address
     @rooms = []
+    @details = {"price" => @price,
+                "address" => @address
+               }
   end
 
   def add_room(room_object)
     @rooms << room_object
+  end
+
+  def above_market_average?
+    if @price > 500000 then true else false end
+  end
+
+  def rooms_from_category(category)
+    sorted_by_category = []
+    @rooms.each do |room|
+      if room.category == category
+        sorted_by_category << room
+      end
+    end
+    sorted_by_category
+  end
+
+  def area
+    total_area = 0
+    @rooms.each do |room|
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,49 @@
+class House
+  attr_reader :price, :address, :rooms, :details
+
+  def initialize(price, address)
+    @price = price[1..-1].to_i
```
aziobrow, line 5: nice job with this

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+    @address = address
+    @rooms = []
+    @total_area = 0
+    @details = {"price" => @price,
```
aziobrow, line 9: this is okay here, but as a matter of best practice, we don't create instance variables unless we have a good reason.  This hash can be created in a method with the same outcome, so we'd prefer to see it that way.

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+                "address" => @address
+               }
+  end
+
+  def add_room(room_object)
+    @rooms << room_object
```
aziobrow, line 15: 👍

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def above_market_average?
+    if @price > 500000 then true else false end
```
aziobrow, line 19: as a matter of convention, you would not see a single line if/else/end like this.  If you're going to do it single line, you'd see it as a ternary. 

Technically, you don't need the explicit true/false. I'd do this:

```
def above_market_average?
   @price > 500000
end
```

Since that statement is a comparison, it returns the correct boolean itself, without needing an explicit true/false return.

aziobrow: Ruby ternary, if you're interested: https://www.rubyguides.com/2019/10/ruby-ternary-operator/

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def rooms_from_category(category)
+    rooms_in_category = []
+    @rooms.each do |room|
+      if room.category == category
+        rooms_in_category << room
+      end
+    end
+    rooms_in_category
```
aziobrow, line 29: this works just fine, and is a good solution. `find_all` is a better tool in this case though, and can be done in a single line. 

```
   @rooms.find_all {|room| room.category == category}
end
```

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def area
+    @rooms.sum {|room| room.area}
```
aziobrow, line 33: very nice

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def price_per_square_foot
+    (@price / area.to_f).round(2)
```
aziobrow, line 37: nice job using the return value of another method here

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by(&:area).reverse!
```
aziobrow, line 41: elegant solution here.  In general, we try to avoid permanently altering the original collection unless we have a very good reason, which means we steer clear of the `!` operator. This would work just fine with the non-destructive form of `reverse`.

lib/house.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def rooms_by_category
+    @rooms.sort_by {|key, value|}.each do |room|
+      room.category
```
aziobrow, line 46: This isn't quite it-- check out `group_by`: http://gregpark.io/blog/ruby-group-by/

lib/room.rb
```diff
@@ -1,12 +1,22 @@
-class Rom
+class Room
+  attr_reader :category
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
-    @width = width
+    @length = length.to_i
```
aziobrow, line 6: length is already an integer when it's passed in, so no particular need to convert to an integer here.

lib/room.rb
```diff
@@ -1,12 +1,22 @@
...
     @length * @width
   end
+
+  def is_painted?
+    @is_painted
+  end
+
+  def paint
+    @is_painted = true
+  end
```
aziobrow, line 21: 👍  Nice job with this class

test/house_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def test_house_can_add_rooms
+    assert_equal [], @house.rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    assert_equal 2, @house.rooms.size
+    assert_instance_of Room, @house.rooms[0]
+    assert_instance_of Room, @house.rooms[1]
+    assert_equal :bedroom, @house.rooms[0].category
```
aziobrow, line 33: probably best to just assert the array-- it saves you time and is usually the most robust expectation.  Then, you just need this single assertion to replace the 4 that you wrote:

`assert_equal [@room1, @room2], @house.rooms`

test/house_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def test_if_house_is_above_market_average
+    refute @house.above_market_average?
```
aziobrow, line 37: If you have multiple possible outcomes for a method, you should test all of them. I'd expect to see a second house here that is above market average, and the corresponding assertion.

test/house_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def test_rooms_from_category
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal 2, @house.rooms_from_category(:bedroom).size
+    assert_equal :bedroom, @house.rooms_from_category(:bedroom).first.category
+    assert_equal :basement, @house.rooms_from_category(:basement).first.category
```
aziobrow, line 48: same feedback as above-- just assert the arrays themselves. 

```
assert_equal [@room1, @room2], @house.rooms_from_category(:bedroom)
assert_equal [@room3], @house.rooms_from_category(:living_room)
assert_equal [@room4], @house.rooms_from_category(:basement)
```

test/house_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def test_rooms_sorted_by_category
+    skip
+    @house.add_room(@room_4)
+    @house.add_room(@room_1)
+    @house.add_room(@room_3)
+    @house.add_room(@room_2)
+
+    assert_equal Hash[:bedroom, @room_1], @house.rooms_by_category
```
aziobrow, line 88: I know we haven't talked about hashes yet, but this assertion doesn't match the expectations laid out in the interaction pattern. I recommend directly copying from the interaction pattern to the test, so that you can be sure that you're really expecting exactly what the interaction pattern is expecting.

test/room_test.rb
```diff
@@ -9,16 +9,22 @@ def test_it_exists
...
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_if_the_room_is_painted_and_can_be_painted
+    room = Room.new(:bedroom, 10, "13")
+    
+    refute room.is_painted?
+    room.paint
+    assert room.is_painted?
```
aziobrow, line 28: 👍

aziobrow (commented):
Really good work, David! The only method you missed was `rooms_by_category`, which was the last one for a reason 🙂  You're right on track in terms of how much knowledge you were able to demonstrate in this independent challenge.  Keep up the good work!

I left inline comments with suggestions/additional detail.  Please let me know if you have any questions!

PR #10, opened 2020-02-04
Jack Puchalla

JPuchalla:
I tend to work at a relatively slow pace, initially, had some difficulties trying to get iteration 1's "def test_it_can_be_painted" to pass but I got it working. I also saw a few places where I could have used setup for instance variables but was practically done with the test at the point.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,2 @@
+class House
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,2 +1,16 @@
 class House
+attr_reader :price, :address, :rooms
+  def initialize(price, address, rooms)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
 end
+
+
+# didnt get very far was working at a pretty slow pace
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,12 +1,21 @@
-class Rom
-
+class Room
+attr_reader :category
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
     @width = width
+    @paint = false
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_it_painted?
+    @paint
+  end
+
+  def paint_item
+    @paint = true
   end
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -1,0 +1,11 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/house.rb'
+require './lib/room.rb'
+
+class HouseTest < Minitest::Test
+
+end
+
+
+# didnt get very far was working at a pretty slow pace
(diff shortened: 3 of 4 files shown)


test/house_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/house.rb'
+require './lib/room.rb'
+
+class HouseTest < Minitest::Test
```
timomitchel, line 6: We need you to be writing the tests BEFORE you implement functionality. You will not be able to pass this mod by writing tests after functionality. This pattern should be discontinued immediately.

lib/room.rb
```diff
@@ -1,12 +1,21 @@
...
     @width = width
+    @paint = false
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_it_painted?
```
timomitchel, line 14: Make sure to name methods exactly what they are named in the README/spec. You will lose points because if you do not because our tests expect them to be named exactly the way they are written in the README.

test/room_test.rb
```diff
@@ -3,22 +3,45 @@
...
     room = Room.new(:bedroom, 10, '13')
     assert_equal :bedroom, room.category
   end
 
   def test_it_can_get_area
-    skip
+
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
```
timomitchel, line 23: This test is failing, because there is no `attr_reader` for `area`. You need to be running your tests and letting the errors guide implementation!

test/room_test.rb
```diff
@@ -3,22 +3,45 @@
...
     assert_equal 180, room2.area
   end
+
+  def test_is_painted
```
timomitchel, line 27: This is a great test.

test/room_test.rb
```diff
@@ -3,22 +3,45 @@
...
+    room1 = Room.new(:bedroom, 10, '13')
+    room2 = Room.new(:living_room, 15, '12')
+
+    assert_equal false, room1.is_it_painted?
+    assert_equal false, room2.is_it_painted?
+  end
+
+  def test_it_can_be_painted
```
timomitchel, line 35: This is another solid test.

lib/house.rb
```diff
@@ -0,0 +1,16 @@
+class House
+attr_reader :price, :address, :rooms
```
timomitchel, line 2: This is a good start, but once again, if there are no tests for this, then in our eyes it's worthless. Make sure to write tests first!

lib/house.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(price, address, rooms)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
```
timomitchel, line 9: This is exactly the implementation that we want here, but without a test it is sad and lonely implementation.

PR #11, opened 2020-02-04
Margo

margoflewelling:


timomitchel:
Would love to see proper commit messages next time `Verb - description of what was changed`

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    rooms << room
+  end
+
+  def above_market_average?
+    if @price[1..-1].to_i > 500000
```
timomitchel, line 15: Nice job with the range to get rid of the `$`. This should have been done in the initialize method, so that price is stored as an integer every time an object is instantiated. This way it's formatted for calculations from that point onward. 

This method could also be simplified to `@price[1..-1].to_i > 500000` which would return the correct boolean.

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category)
```
timomitchel, line 22: Great job using each to get the job done here! You can save a few lines with `find_all`: 

`@rooms.find_all {|room| room.category == category}`

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+      if room.category == category
+        rooms_category << room
+      end
+    end
+    rooms_category
+  end
+
+  def area
+    all_room_areas = []
+    @rooms.each do |room|
```
timomitchel, line 34: Fantastic job using a combination of each and sum to get the job done in this method. `sum` can take a block to make this even cleaner: 

`@rooms.sum {|room| room.area}`

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+      all_room_areas << room.area
+    end
+    all_room_areas.sum
+  end
+
+  def details
+    details = {price: @price[1..-1].to_i, address: @address}
```
timomitchel, line 41: Anytime you find yourself duplicating work ie: `@price[1..-1].to_i` then there is definitely a way to clean that up and only do the work once. In this case, right in `initialize`. Try to keep the code DRY (Don't repeat yourself).

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def price_per_square_foot
+    (@price[1..-1].to_i / area.to_f).round(2)
+  end
+
+  def rooms_sorted_by_area
+      @rooms.sort_by { |room| -room.area }
```
timomitchel, line 49: Clean and effective.

lib/house.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def rooms_by_category
+      @rooms.sort_by { |room| room.category }
```
timomitchel, line 53: We actually need to build a hash here. This method is certainly the most difficult, but we could accomplish it with something like: 

```
rooms_by_categories = {}
@rooms.each do |room|
if rooms_by_categories[room.category]
rooms_by_categories[room.category] << room  
else  
rooms_by_categories[room.category] = [room]    
end
end
```

test/house_test.rb
```diff
@@ -0,0 +1,105 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/room'
+require './lib/house'
+
+class HouseTest < Minitest::Test
+
+  def test_it_exists_and_has_attrs
```
timomitchel, line 8: This should be split into two tests. One for exists and one for attrs.

test/house_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+    assert_equal "$400000", house.price
+    assert_equal "123 sugar lane", house.address
+    assert_equal [], house.rooms
+  end
+
+  def test_adding_rooms
+    house = House.new("$400000", "123 sugar lane")
```
timomitchel, line 17: You can save yourself some time by utilizing a `setup` method to create your objects.

test/house_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
```
timomitchel, line 40: Great job with some robust testing here testing two categories.

test/house_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+    assert_equal 1900, house.area
+  end
+
+
+  def test_house_details
+    house = House.new("$400000", "123 sugar lane")
+    hash = {:price => 400000, :address => "123 sugar lane"}
```
timomitchel, line 60: We don't want to use data types (hash, array, etc) as names for our variables or methods. `expected` is a better name for the variable here that follows convention.

test/house_test.rb
```diff
@@ -0,0 +1,105 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_4)
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    rooms_by_size = [room_4, room_3, room_2, room_1]
```
timomitchel, line 87: Much better naming here.

test/house_test.rb
```diff
@@ -0,0 +1,105 @@
...
+     room_1 = Room.new(:bedroom, 10, '13')
+     room_2 = Room.new(:bedroom, 11, '15')
+     room_3 = Room.new(:living_room, 25, '15')
+     room_4 = Room.new(:basement, 30, '41')
+     house.add_room(room_4)
+     house.add_room(room_1)
+     house.add_room(room_2)
+     house.add_room(room_3)
+     category_hash = {:bedroom => [room_1, room_2], :living_room => [room_3], :basement => [room_4]}
+     assert_equal category_hash, house.rooms_by_category
```
timomitchel, line 102: Really solid test setup here other than the naming of `category_hash`!

timomitchel (commented):
Overall nice job on this challenge Margo! You're looking comfortable with each and with TDD which are incredibly important parts of Mod-1! This meets our expectations!

PR #13, opened 2020-02-04
Whitney Kidd

whitneykidd:


test/house_test.rb
```diff
@@ -0,0 +1,80 @@
...
+require "./lib/room"
+require "./lib/house"
+
+class HouseTest < Minitest::Test
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_instance_of House, house
+  end
+
```
memcmahon, line 12: I would have also expected to see tests for `address` and `price`

lib/house.rb
```diff
@@ -0,0 +1,44 @@
+
+
+class House
+  attr_reader :price, :address, :rooms
```
memcmahon, line 4: Keep an eye on the interaction pattern - the price should have been a separate method that returned price as an integer (`400000` instead of `"$4000000"`)

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def add_room(new_room)
+      @rooms << new_room
+    end
+
+    def above_market_average?
+      if @price.to_i > 500000
+        true
+      else
```
memcmahon, line 18: An `else` does not take a conditional - the way this should be written is like this:

```ruby
if @price.to_i > 500000
  true
elsif @price < 500000
  false
end
```

OR

```ruby
if @price.to_i > 500000
  true
else
  false
end
```

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+      @address = address
+      @rooms = []
+    end
+
+    def add_room(new_room)
+      @rooms << new_room
+    end
+
+    def above_market_average?
+      if @price.to_i > 500000
```
memcmahon, line 16: This method, as written won't work quite as expected, because if a House is created correctly as `House.new("$400000", "some address")`, `@price.to_i` will return `0`

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+        true
+      else
+        @price.to_i < 500000
+        false
+      end
+    end
+
```
memcmahon, line 23: Keep an eye on the indentation from here to the end of the class - as written, the class looks like this:

```ruby
class House

  attr_reader ...
  
  def initialize(...)
  end
  
  def add_room(...)
  end

  def above_market_average?
  end

  def rooms_from_category(...)
    def area
    end
    
    def details
    end
  end
end
```

with the `area` and `details` *inside* the `rooms_from_category method`, which causes some unexpected behavior.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+    def rooms_from_category(category)
+      rooms_from_category = []
+      @rooms.select do |room|
+        if room.category == category
+          rooms_from_category << room
+        end
+    end
+
+    def area
```
memcmahon, line 32: This method looks great!  But see my note about indentation!

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+     total_area = 0
+     rooms.each do |room|
+       total_area += room.area
+     end
+     total_area
+   end
+
+    def details
```
memcmahon, line 40: The logic and how this hash is built is correct, but because `price` isn't working as expected, this also does not quite give us what we are looking for.

lib/house.rb
```diff
@@ -0,0 +1,44 @@
...
+        true
+      else
+        @price.to_i < 500000
+        false
+      end
+    end
+
+    def rooms_from_category(category)
+      rooms_from_category = []
+      @rooms.select do |room|
```
memcmahon, line 26: Good use of select, but the syntax is a bit off - it should look like this:

```ruby
  @rooms.select do |room|
    room.category == category
  end
end
```

test/room_test.rb
```diff
@@ -4,21 +4,39 @@
...
     assert_equal 180, room2.area
   end
+
+  def test_it_returns_painted_false
+
+    room = Room.new(:bedroom, 10, '13')
+    assert_equal false, room.is_painted?
+  end
+
+  def test_it_returns_painted_true
```
memcmahon, line 32: Great job building and testing these paint related methods!

memcmahon (commented):
Hey Whitney - there is some great stuff here, but not quite where we would like you to be. Make sure to keep a close eye on the interaction pattern next time, and I would encourage you to work with your mento on cleaning this up, and finishing it :)

PR #16, opened 2020-02-04
Mariana Cid

Mariana-21:


lib/room.rb
```diff
@@ -1,12 +1,22 @@
...
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    @is_painted
+  end
+
+  def paint
+    @is_painted = true
   end
```
aziobrow, line 21: Nice job with this room class!

lib/house.rb
```diff
@@ -0,0 +1,42 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
```
aziobrow, line 5: this isn't quite right-- the price is a string when it gets passed in, so we'd need to convert 

`@price = price.to_i` to get this to work the way the interaction pattern expects

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+    @address = address
+    @rooms = []
+    @above_market_average = false
```
aziobrow, line 8: this doesn't need to be an instance variable-- you can just have this be a method

```
def above_market_average
     @price > 500_000
end
```

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+    @sum = 0
+    @area_room = []
+    @detail_hsh = {}
```
aziobrow, line 11: You don't need any of these instance variables either-- everything you want to do can be accomplished with methods

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
```
aziobrow, line 15: 👍

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def above_market_average?
+    @above_market_average
```
aziobrow, line 19: since you set it to a default value of false in `def initialize`, there's no way a house can ever be above market average.  This should be `@price > 500000` instead (if the price is greater than 500000, this statement will return true, or else, it will return false.

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.map do |room|
+      if room.category == category
+        room
+      end
+    end.compact
```
aziobrow, line 27: Nice! Great job with this one

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def area
+    @rooms.map do |room|
+      room.length * room.width.to_i
```
aziobrow, line 32: Close! You'd want to use a `sum` here, instead of a `map`.

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+      # require "pry"; binding.pry
+    end
+  end
+
+  def details
+    @detail_hsh  = {
+                      "price" => "$400000",
+                      "address" => "123 sugar lane"}
```
aziobrow, line 40: You don't need the `@detail_hsh` here. You can just have this method return the has you want it to have.  Also, this is hardcoded--the way you have this written, every house would have a price of $400000 and an address of "123 sugar lane" for its details, which probably shouldn't be the case. 

```
def details
   info = {}
   info["price"] = "$" + @price.to_i
   info["address] = @address 
end 
```

test/house_test.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+
+    assert_equal [room_1, room_2], house.rooms
+  end
+
+  def test_its_above_market_average
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal false, house.above_market_average?
```
aziobrow, line 33: This test passes, but only because you set it to a default value of false.  You'd have a better test if you also added another house with a value of greater than 500000 and make sure that it is above_market_average.

test/house_test.rb
```diff
@@ -0,0 +1,57 @@
...
+    room_4 = Room.new(:basement, 30, '41')
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+    assert_equal [room_4], house.rooms_from_category(:basement)
+    # assert_equal 1900, house.area
```
aziobrow, line 50: I would expect to see this area assertion in a different test, since this is a different method.

test/house_test.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def test_it_shows_details
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal ({"price" => "$400000", "address" => "123 sugar lane"}), house.details
```
aziobrow, line 55: Same deal here-- this test may pass, but only because you hardcoded the values. This wouldn't pass if your house had a different price or address.

test/room_test.rb
```diff
@@ -9,16 +9,23 @@ def test_it_exists
...
   end
+
+  def test_if_room_is_painted
+    room = Room.new(:bedroom, 10, '13')
+    assert_equal false, room.is_painted?
+
+    room.paint
+
+    assert_equal true, room.is_painted? 
+  end
```
aziobrow, line 30: this is a nice test! I like the progression here.

aziobrow (commented):
Hey Mariana--

If you want to go back through this, the methods that you didn't get working were: price, above_market_average?, area, details, price_per_square_foot, rooms_sorted_by_area, rooms_by_category. Some of them, you just didn't have time to get to, and others, you didn't get working in exactly the way that the interaction pattern wanted. 

You did show some good work, which you should be proud of, but you're not quite where we'd like you to be at this point-- the goal for these is to completely and accurately finish Iteration 3. Let me know if you'd like to pair or if you have any questions!

PR #17, opened 2020-02-04
Michelle Foley

foleymichelle9:


lib/room.rb
```diff
@@ -1,12 +1,25 @@
-class Rom
+class Room
+  attr_reader :category, :length, :width
```
memcmahon, line 2: Since we don't ever call `length` or `width` as methods, you don't need them as attr_readers

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
+    @width = width.to_i
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+
+  def paint
```
memcmahon, line 15: The logic that you have for these paint methods looks like it is on the right track, but a better solution would look something like this:

```ruby
class Room
  ...
    ...
    @is_painted = false
  end

  ...

  def paint
    @is_painted = true
  end

  def is_painted?
    @is_painted
  end
end
```

test/room_test.rb
```diff
@@ -9,16 +9,20 @@ def test_it_exists
...
 
   def test_it_can_get_area
-    skip
     room1 = Room.new(:bedroom, 10, '13')
     room2 = Room.new(:living_room, 15, '12')
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_it_has_paint
```
memcmahon, line 23: Great job on this test!

memcmahon (commented):
Hey Michelle - this is not quite where we would like you to be. I would encourage you to work with your mentor to review and complete this challenge - there are a lot of key concepts covered here, so it would be a great exercise!

test/room_test.rb
foleymichelle9: @memcmahon Thank you for the feedback! I'll work with my mentors to get it done :)

PR #18, opened 2020-02-04
Colin Alexander

colintalex:


lib/house.rb
```diff
@@ -0,0 +1,32 @@
+class House
+  attr_reader :price, :address, :rooms, :house_area
```
memcmahon, line 2: Keep an eye on the interaction pattern, we were really looking for a `price` method that would return the price of the house as an integer - this would also help with a few methods below

lib/house.rb
```diff
@@ -0,0 +1,32 @@
...
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+    @house_area = 0
+  end
+
+  def details
+    {"price" => @price.tr('$', '').to_i, "address" => @address}
```
memcmahon, line 11: Great job with this method!

lib/house.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def area
```
memcmahon, line 18: The logic here is great, but you don't really need an instance variable for `@house_area`; you could format the method like this:

```ruby
def area
  house_area = 0
  @rooms.each do |room|
    house_area += room.area
  end
  house_area
end
```

lib/house.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+    @house_area
+  end
+
+  def above_market_average?
+    @price.tr('$', '').to_i >= 500000
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all{ |room| room.category == category}
```
memcmahon, line 30: Great use of find_all!

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
     @width = width
+    @coats_of_paint = 0
   end
 
   def area
-    @length * @width
+    @length.to_f * @width.to_f
+  end
+
+  def painted?
```
memcmahon, line 14: Keep a close eye on the interaction pattern!  The method should be called `is_painted?`

test/house_test.rb
```diff
@@ -0,0 +1,75 @@
...
+
+class HouseTest < Minitest::Test
+def test_it_exists
+  house = House.new("$400000", "123 sugar lane")
+  assert_instance_of House, house
+end
+
+def test_it_has_price
+  house = House.new("$400000", "123 sugar lane")
+  assert_equal "$400000", house.price
```
memcmahon, line 14: Keep a *close* eye on the interaction pattern; we wanted this method to return an integer

memcmahon (commented):
Hey Colin - all of the logic here looks good, but a few of the methods don't quite follow the interaction pattern.  We do expect that when we ask for a method to return a specific datatype, or be called something specific, that the methods that you create follow those conventions!

PR #19, opened 2020-02-04
Javier Aguilar

javier-aguilar:


lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+  attr_reader :address, :rooms
+
+  def initialize(price_paramater, address_parameter)
+    @price = price_paramater
+    @address = address_parameter
+    @rooms = []
+  end
+
+  def price
+    @price.gsub("$", '').to_i
```
aziobrow, line 11: nice job writing this method yourself instead of relying on the attr_reader, since you needed to do some custom work here

lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    price > 500000 ? true : false
```
aziobrow, line 19: This is good, although you don't need the ternary here at all. 

```
def above_market_average?
   price > 500000
end
```

`price > 500000` is a comparison and thus returns the correct boolean all by itself, so no need for an explicit true/false

lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.select {|room| room.category == category}
```
aziobrow, line 23: this is a very good enumerable for this job

lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def area
+    total_area = 0
+    @rooms.each {|room| total_area += room.area}
+    total_area
```
aziobrow, line 29: This works just fine-- sum is also an enumerable that could have worked really nicely here. 

```
def area 
   @rooms.sum {|room| total_area += room.area}
end
```

lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def details
+    details = {"price" => price, "address" => @address}
+    details
```
aziobrow, line 34: This is also good, although you don't need to set it equal to a variable at all. 

```
def details
     {"price" => price, "address" => @address}
end
```

lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def price_per_square_foot
+    total_per_square_foot = 0
+    if area != 0
+      total_per_square_foot = (price / area.to_f).round(2)
+    end
+    total_per_square_foot
```
aziobrow, line 42: good job accounting for the 0 here, although if you do that, i would provide an alternative option if area does == 0, as a matter of best practice.  Nice job using the return value of another method as a helper here.

lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+    rooms.each do |room|
+      if rooms_sorted.first == nil
+        rooms_sorted.append room
+      elsif room.area > rooms_sorted.first.area
+        rooms_sorted.prepend room
+      else
+        rooms_sorted.append room
+      end
+    end
+    rooms_sorted
```
aziobrow, line 56: This doesn't quite work... which is understandable, as there's a whole universe of algorithmic challenges around building sort methods... fortunately, you don't have to worry about that. Ruby has a nice, built-in sort method for you. (defaults to ascending order, least to greatest)

```
def rooms_sorted_by_area
   @rooms.sort_by {|room| room.area}.reverse
end
```
(reversed to switch the order to descending, as per the interaction pattern)

lib/house.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def rooms_by_category
+    # iterate through rooms and if they are in the same category put in the same array
+    # if in a different category put in a seperate array
+    # ex:
+    # {:bedroom=>[#<Room:0x00007fccd29b5720...>, #<Room:0x00007fccd2985f48...>]}
```
aziobrow, line 63: I like the pseudocode here

lib/house.rb
javier-aguilar: Ugh I keep forgetting about .sum 😞

lib/room.rb
```diff
@@ -1,12 +1,23 @@
...
-  def initialize(category, length, width)
-    @category = category
-    @length = 10
-    @width = width
+  def initialize(category_parameter, length_parameter, width_parameter)
+    @category = category_parameter
+    @length = length_parameter.to_i
+    @width = width_parameter.to_i
+    @painted = false
+    @area = 0
```
aziobrow, line 9: No need for this to be an instance variable. This can just be calculated by the method when the method is called.  It's best practice to avoid creating instance variables unless we really need them.

lib/room.rb
```diff
@@ -1,12 +1,23 @@
...
+class Room
+  attr_reader :category
 
-  def initialize(category, length, width)
-    @category = category
-    @length = 10
-    @width = width
+  def initialize(category_parameter, length_parameter, width_parameter)
+    @category = category_parameter
+    @length = length_parameter.to_i
```
aziobrow, line 6: length is already an integer, so no need to convert here

javier-aguilar:
Thanks for all the feedback @aziobrow. These are all very helpful!!!

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_has_a_price
+    house1 = House.new("$400000", "123 sugar lane")
+    assert_equal 400000, house1.price
+
+    house2 = House.new("250000", "124 sugar lane")
+    assert_equal 250000, house2.price
```
aziobrow, line 18: This is fine that you tested two different houses, but unless there's something substantially different about the two of them, there's really no reason to do so.

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  def test_it_has_an_address
+    house1 = House.new("$400000", "123 sugar lane")
+    assert_equal "123 sugar lane", house1.address
+
+    house2 = House.new("250000", "999 Other Way")
+    assert_equal "999 Other Way", house2.address
```
aziobrow, line 26: same here

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    house.add_room(room2)
+    assert_equal [room1, room2],house.rooms
+  end
+
+  def test_if_it_is_above_market_average
+    house1 = House.new("$600000", "123 ABC Lane")
+    assert_equal true, house1.above_market_average?
+
+    house2 = House.new("$400000", "123 sugar lane")
+    assert_equal false, house2.above_market_average?
```
aziobrow, line 46: here, it's good to test two houses, since they give you different outcomes from the method you're testing. good work.

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    house.add_room(room3)
+    house.add_room(room4)
+    assert_equal 1900, house.area
+  end
+
+  def test_it_has_details
+    house = House.new("$400000", "123 sugar lane")
+    details = {"price" => 400000, "address" => "123 sugar lane"}
+
+    assert_equal details, house.details
```
aziobrow, line 86: 👍

test/house_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    room3 = Room.new(:living_room, 25, '15')
+    room4 = Room.new(:basement, 30, '41')
+
+    house.add_room(room1)
+    house.add_room(room2)
+    house.add_room(room3)
+    house.add_room(room4)
+
+    # require "pry"; binding.pry
+    assert_equal [], house.rooms_by_category
```
aziobrow, line 138: nice job writing your tests first

test/room_test.rb
```diff
@@ -9,16 +9,22 @@ def test_it_exists
...
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_it_can_be_painted
+    room = Room.new(:bedroom, 10, '13')
+    assert_equal false, room.is_painted?
+
+    room.paint
+    assert_equal true, room.is_painted?
```
aziobrow, line 28: 👍

aziobrow (commented):
Very nice work on this, Javi! You did a great job with both implementation and following the interaction pattern.  The only methods you missed were `rooms_sorted_by_area` and `rooms_by_category` . Please see my inline comments for more detail.  

PR #20, opened 2020-02-04
Chris Stankus

chrisstankus99:


lib/room.rb
```diff
@@ -1,12 +1,22 @@
...
     @length * @width
   end
+
+  def is_painted?
+    @painted
+  end
+
+  def paint
+    @painted = true
+  end
```
aziobrow, line 21: very nice work with this class! Good job working with the boolean method and catching that you needed to convert the width to an integer.

lib/house.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  attr_reader :price, :address, :rooms
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
```
aziobrow, line 11: 👍

lib/house.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def rooms_from_category(room_category)
+    @rooms.find_all {|room| room.category == room_category}
```
aziobrow, line 19: beautiful! this is exactly the right enumerable for this job.

lib/house.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def area
+    area = 0
+    @rooms.each {|room| area = area + room.area}
+    area
```
aziobrow: Great! This works well.  Ruby has shorthand for adding and reassigning a value, so you could have written your line 20 like this: 

`@rooms.each {|room| area += room.area}`


Ruby also has an enumerable called `sum` which could have made this whole method just one line. 

`@rooms.sum {|room| room.area}`

lib/house.rb
```diff
@@ -0,0 +1,35 @@
+class House
+
+  attr_reader :price, :address, :rooms
+  def initialize(price, address)
+    @price = price
```
aziobrow: If you look closely at the interaction pattern, price should be an integer. so you'd want to do `@price = price.to_i`

lib/house.rb
```diff
@@ -0,0 +1,35 @@
...
+    @rooms.find_all {|room| room.category == room_category}
+  end
+
+  def area
+    area = 0
+    @rooms.each {|room| area = area + room.area}
+    area
+  end
+
+  def house_details
```
aziobrow: make sure you look at the interaction pattern really carefully-- your logic is correct, but this method should just be called `details`

lib/house.rb
```diff
@@ -0,0 +1,35 @@
...
+    details = {
+      "price" => price,
+      "address" => address
+    }
+    details
+  end
+
+  def price_per_square_foot
+    (price.delete_prefix("$").to_f / area).round(2)
```
aziobrow: don't get too used to `delete_prefix`-- it's part of the very latest and greatest version of ruby, which your place of work will almost certainly not have 🙂  I'd try to find other ways to do this.

test/house_test.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def test_if_house_can_get_rooms_from_category
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal [@room_1, @room_2], @house.rooms_from_category(:bedroom)
+    assert_equal [@room_4], @house.rooms_from_category(:basement)
+    assert_equal 1900, @house.area
```
aziobrow, line 57: I would expect to see this as a different test.  We would not typically put assertions for two different methods in the same test, even if it has the same setup.

test/house_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    assert_equal 210.53, @house.price_per_square_foot
+  end
+
+  def test_it_can_get_rooms_sorted_by_area
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal [@room_4, @room_3, @room_2, @room_1], @house.rooms_sorted_by_area
```
aziobrow, line 80: good job writing your tests first

test/room_test.rb
```diff
@@ -3,22 +3,26 @@
...
-    assert_equal 130, room1.area
-    assert_equal 180, room2.area
+    assert_equal 130, @room_1.area
+    assert_equal 180, @room_2.area
+  end
+
+  def test_is_painted_and_paint
+    refute @room_1.is_painted?
+    @room_1.paint
+    assert @room_1.is_painted?
```
aziobrow, line 27: Setup methods can save time, but in this case, rewriting what was already there to use a setup probably cost you more time than it saved, since you only needed to write one test of your own.  Only do setup methods when they make sense.

aziobrow (commented):
Chris-- overall, you showed some really good work here (and I know we changed up the schedule b/c of the snow day, so we expected you to do some work with hashes that we ended up having not taught yet!) 

If you'd like to go back and look over this, the methods you missed were price, above_market_average, details, rooms_sorted_by_area, rooms_by_category.  Some of those, you just didn't get to (although it looks like you just skipped `above_market_average?`, since that was early in Iteration 3, and others, you missed because what you wrote didn't match the interaction pattern exactly. 

In general, this is close to where we want you to be, although ideally, we do want you to be completely and accurately finishing through iteration 3, so make sure to follow that pattern! 

Let me know if you have any questions or if my comments need any more clarity.

test/house_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal [@room_1, @room_2], @house.rooms_from_category(:bedroom)
+    assert_equal [@room_4], @house.rooms_from_category(:basement)
+    assert_equal 1900, @house.area
+  end
+
+  def test_if_can_get_house_details
+    assert_equal ({"price"=>"$400000", "address"=>"123 sugar lane"}), @house.house_details
```
aziobrow: be sure to follow the interaction pattern exactly.  This method should just be called `details`, not `house_details`

test/house_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    @room_4 = Room.new(:basement, 30, '41')
+    @house = House.new("$400000", "123 sugar lane")
+  end
+
+  def test_it_exists
+    assert_instance_of House, @house
+  end
+
+  def test_it_has_attributes
+    assert_equal "$400000", @house.price
```
aziobrow: be sure to follow the interaction pattern exactly.  it's asking for the price as an integer.

PR #21, opened 2020-02-04
Brian Greeson

brian-greeson:


lib/house.rb
```diff
@@ -0,0 +1,33 @@
...
+  def price
+    @price.delete('$').to_i
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    self.price > 5000000
```
memcmahon, line 18: sooooo close!  But this should be `self.price > 500000` (you've got too many zeros 😉)

lib/house.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def rooms_from_category(category)
+    rooms.select {|room| room.category == category}
```
memcmahon, line 22: Great use of select

lib/house.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def area
+    rooms.map.sum { |room| room.area  }
```
memcmahon, line 26: This does work, but you don't need the `map`, you can use `rooms.sum...`

test/house_test.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def test_it_can_add_rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_it_can_be_below_market_average
+    assert_equal false, @house.above_market_average?
```
memcmahon, line 40: For any method that could return true or false, I would write a test for both (this might have shown your error in the method)

memcmahon (commented):
Hey Brian - this looks really good!  Take a look at my notes, but overall, it looks like you are on track for where we would expect you to be 🙂

PR #23, opened 2020-02-04
Mike Hernandez

mikez321:
Edit Room class to pass tests.
Wrote extra tests for pry workflow.
Created House class and tests to follow PRY.
Merged from my work branch into master.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -26,4 +26,9 @@ class House
     end
   end
 
+  def area
+    room_areas = rooms.map { |room| room.area }
+    room_areas.sum
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -10,4 +10,8 @@ class House
     (@price.delete("$")).to_i
   end
 
+  def add_room(room)
+    @rooms << room
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -31,4 +31,11 @@ class House
     room_areas.sum
   end
 
+  def details
+    details = Hash.new
+    details["price"] = price
+    details["address"] = address
+    details
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -42,4 +42,9 @@ class House
     (price.to_f / area.to_f).round(2)
   end
 
+  def rooms_sorted_by_area
+    room_areas = (rooms.map { |room| room.area }).sort.reverse
+    require "Pry"; binding.pry
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,13 @@
+class House
+  attr_reader :address, :rooms
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def price
+    (@price.delete("$")).to_i
+  end
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -4,6 +4,7 @@ class House
     @price = price
     @address = address
     @rooms = []
+    @market_average = 500000
   end
 
   def price
@@ -13,5 +14,12 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    return true if price > @market_average
+      false
+  end
+
+  def rooms_from_category(category)
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -21,5 +21,9 @@ class House
   end
 
   def rooms_from_category(category)
+    rooms.find_all do |room|
+      category == room.category
+    end
+  end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -38,4 +38,8 @@ class House
     details
   end
 
+  def price_per_square_foot
+    (price.to_f / area.to_f).round(2)
+  end
+
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -1,9 +1,9 @@
-class Rom
-
-  def initialize(category, length, width)
+class Room
+  attr_reader :category
+  def initialize(category, length = 10, width)
     @category = category
-    @length = 10
-    @width = width
+    @length = length.to_i
+    @width = width.to_i
   end
 
   def area
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -4,9 +4,19 @@ class Room
     @category = category
     @length = length.to_i
     @width = width.to_i
+    @is_painted = false
   end
 
   def area
     @length * @width
   end
+
+  def is_painted?
+    @is_painted
+  end
+
+  def paint
+    @is_painted = true
+  end
+
 end
--- a/lib/room.rb
+++ b/lib/room.rb
@@ -19,4 +19,8 @@ class Room
     @is_painted = true
   end
 
+  def area
+    @length * @width
+  end
+
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -17,27 +17,16 @@ class RoomTest < Minitest::Test
     assert_equal "123 sugar lane",house.address
   end
 
-  def rooms_can_be_added
+  def test_rooms_can_be_added
     house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
 
     assert_equal [], house.rooms
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    assert_equal [room_1, room_2], house.rooms
   end
 
-
 end
-
-#house.rooms
-#=> []
-
-#room_1 = Room.new(:bedroom, 10, '13')
-#=> #<Room:0x00007fccd29b5720...>
-
-#room_2 = Room.new(:bedroom, 11, '15')
-#=> #<Room:0x00007fccd2985f48...>
-
-#house.add_room(room_1)
-
-#house.add_room(room_2)
-
-#house.rooms
-#=> [#<Room:0x00007fccd29b5720...>, #<Room:0x00007fccd2985f48...>]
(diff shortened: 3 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,50 @@
+class House
+  attr_reader :address, :rooms
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+    @market_average = 500000
```
aziobrow, line 7: This is okay, but as a matter of best practice, we try to avoid setting instance variables when we don't need to.  Since this is only being used in one method, your `above_market_average?` method, you can just reference that value directly in that method and then it doesn't need to be exposed here.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def price
+    (@price.delete("$")).to_i
```
aziobrow, line 11: nice job catching that you had to write this method yourself instead of relying on an attr_reader

aziobrow: you don't need parentheses here, as the part you have in parentheses is all one method call anyway

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    return true if price > @market_average
+      false
```
aziobrow, line 20: i like this guard clause, although you don't need these explicit true/falses here. 

```
def above_market_average?
   price > @market_average
end
```

Since `price > @market_average` is a comparison, it already returns the correct boolean value as a result, so that's really all you need here.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def rooms_from_category(category)
+    rooms.find_all do |room|
+      category == room.category
```
aziobrow, line 25: beautiful. perfect enumerable for this work.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
+
+  def area
+    room_areas = rooms.map { |room| room.area }
+    room_areas.sum
```
aziobrow, line 31: this works fine, but `sum` is actually an enumerable too.  So the best solution here would be:

```
def area 
   rooms.sum { |room| room.area }
end
```

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def details
+    details = Hash.new
+    details["price"] = price
+    details["address"] = address
+    details
```
aziobrow, line 38: very nicely done

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def price_per_square_foot
+    (price.to_f / area.to_f).round(2)
```
aziobrow, line 42: nice job using the return value of another method here to avoid duplicating work. 

Small point, but you only have to convert one value to a float in order for the result to be a float, so no need to do `.to_f` on both values.

lib/house.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def rooms_sorted_by_area
+    room_areas = (rooms.map { |room| room.area }).sort.reverse
```
aziobrow, line 46: This doesn't work quite right, since you're mapping for the areas of the different rooms, you're going to get an array of integers instead of an array of rooms, which is not what you want.   `sort` or `sort_by` (fundamentally very similar) takes a block in which you can give instructions as to how to sort. 

```
def rooms_sorted_by_area
    rooms.sort_by {|room| room.area}.reverse
end
```

lib/room.rb
```diff
@@ -1,12 +1,26 @@
-class Rom
-
-  def initialize(category, length, width)
+class Room
+  attr_reader :category
+  def initialize(category, length = 10, width)
```
aziobrow, line 3: I don't think this makes sense as a default value here-- there's nothing logical to suggest that most rooms have a length of 10.

lib/room.rb
```diff
@@ -1,12 +1,26 @@
...
     @category = category
-    @length = 10
-    @width = width
+    @length = length.to_i
```
aziobrow, line 5: length is already an integer when it's passed in, so no need to convert to_i here.

lib/room.rb
```diff
@@ -1,12 +1,26 @@
...
+  end
+
+  def paint
+    @is_painted = true
   end
 
   def area
     @length * @width
   end
+
```
aziobrow, line 25: nicely done on this class

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+    assert_equal [room_1, room_2], house.rooms
+  end
+
+  def test_it_can_be_above_or_below_market_average
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house.above_market_average?
```
aziobrow, line 35: if you have a method that can return multiple possible outcomes, test them all.  I'd expect to see another house here that is above market average, and an assertion to go with that.

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal 1900, house.area
+  end
+
+  def test_show_details
+    house = House.new("$400000", "123 sugar lane")
+    house.details
```
aziobrow, line 69: this doesn't need to be here-- you're calling `house.details` in your assertion on line 73, so this is just calling it an extra time for no particular reason.  Although it's unnecessary here, it doesn't hurt you, but in some cases, it can really mess up your results, so best to just avoid it altogether.

test/house_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    house.rooms_sorted_by_area
+
+    assert_equal [room_4, room_3, room_2, room_1], house.rooms_sorted_by_area
+
```
aziobrow, line 105: 👍

aziobrow (commented):
Overall, very nice work.  The only two methods you missed/didn't get to were `rooms_sorted_by_area` and  `rooms_by_category`, but you made it accurately and completely through Iteration 3, which is what we were hoping for.  I left some inline comments with some specifics/suggestions.

aziobrow:
Also, I appreciate you following the practice of leaving good PR comments when you make one!

lib/house.rb
mikez321: So hard coding it was the ideal solution?  I put it in a variable so it could be changed in one place and referenced wherever it was needed.

lib/house.rb
mikez321: I'm really bad about that and do it all the time... It's just the way I think I guess.  I'll try to work on it.

lib/house.rb
mikez321: Good to know!

lib/house.rb
mikez321: I was saved by some of the Ruby Exercises that forced me to work with hashes.  I don't feel comfortable with them yet, but I had enough exposure that I didn't totally freak out!

lib/house.rb
mikez321: I was running out of time and didn't get a chance to finish that up.  After I saw the result in PRY I realized I was going down the wrong path, but instead of continuing I went ahead and merged my "work" branch and made my PR at 10:59!

lib/room.rb
mikez321: I set it as a default because the starting code had it hard-coded in the initialize method so I thought that was intentional.  This preserved the default setting but also allowed the room to be a different size and pass other tests.  But totally agree... no reason a room should have a length of 10.


class Rom

    @category = category
    @length = 10
    @width = width
  end

lib/room.rb
mikez321: Length was passed in as a string in the first test so I did that to allow string or integer input.

  def test_it_exists
    room = Room.new(:bedroom, 10, '13')
    assert_instance_of Room, room
  end

test/house_test.rb
mikez321: I did add a test for the additional area room method I made, but I missed this one!  Productive struggle!  :)

test/house_test.rb
mikez321: I was PRYing at that point and had it there to hit a pry.  Forgot to take it out.  I'm sure one of these days I'm going to destroy my code doing that and not know why.  I'll remember your warning on that day!

lib/house.rb
aziobrow: Your solution is hardcoded too-- it's a value that's explicitly being set to a single value (as opposed to being passed in via an argument).  

For your use case, there's no reason to expose that information to the entire class when you only need it in one method. You should always choose to give things the smallest scope that meets your needs as a matter of security.  

Since you're only using that information in one place currently, if you needed to change it, you'd still go to that one place and change it.  The only difference is whether that one place is in `def initialize` or in the `above_market_average?` method.  If you found yourself needing that information in multiple methods, then you'd consider an instance variable.

lib/house.rb
aziobrow: it's not a big deal; just something to have floating around in your head 🙂

lib/house.rb
aziobrow: yeah... ideally, we would have taught hashes before this independent challenge, but... ❄️

lib/room.rb
aziobrow: i see where you're coming from.  it was intentional in the sense that we wanted you to see that it was causing problems and that it needed to be set equal to the length in the parameter, but there was no particular meaning for the 10 specifically.

lib/room.rb
aziobrow: The second argument, the length, is 10, which is an integer and thus didn't need to be converted.  The third argument, width, was a string, and needed to be converted.

PR #24, opened 2020-02-04
Ashkan Abbasi

Ashkanthegreat:
I couldn't get passed this error: NoMethodError: undefined method `price' for nil:NilClass
    test/house_test.rb:15:in `test_it_has_price_address_rooms'

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -6,9 +6,13 @@ class House
   attr_reader :price, :address, :rooms
 
   def initialize(price, address, rooms)
-    @price = price 
+    @price = price.delete('$').to_i
     @address = address
     @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room 
   end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,5 @@
+class House
+  def initialize(price, address)
+  end  
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,5 +1,14 @@
+require 'minitest'
+require './lib/room'
+
+
 class House
-  def initialize(price, address)
-  end  
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address, rooms)
+    @price = price 
+    @address = address
+    @rooms = []
+  end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -12,7 +12,11 @@ class House
   end
 
   def add_room(room)
-    @rooms << room 
+    @rooms << room
+  end
+
+  def above_market_average?
+    @price > 500000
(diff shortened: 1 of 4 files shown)


test/house_test.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class HouseTest < Minitest::Test
+
+  def test_it_exists
+    house = House.new('$400000', '123 sugar lane', [])
+    assert_instance_of House, house
+  end
+
+  def test_it_has_price_address_rooms
+    assert_equal '$400000', @house.price
```
timomitchel: So the reason that you were seeing `NoMethodError: undefined method price' for nil:NilClass test/house_test.rb:15:in ` is because `@house` was never defined in your test. For each method in your test you need to create a new instance of `House` just like you did on line 10 of this file in `test_it_exists`. That should be the starting point of every test method. Let me know if you have more questions about this.

test/room_test.rb
```diff
@@ -1,24 +1,26 @@
...
 
   def test_it_can_get_area
-    skip
-    room1 = Room.new(:bedroom, 10, '13')
-    room2 = Room.new(:living_room, 15, '12')
+    room1 = Room.new(:bedroom, 10, 13)
+    room2 = Room.new(:living_room, 15, 12)
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
```
timomitchel, line 30: We would have liked to see you add tests in this file to test the first iteration of the interaction pattern.

lib/house.rb
```diff
@@ -0,0 +1,14 @@
+require 'minitest'
+require './lib/room'
```
timomitchel, line 2: We never need to require minitest in a class file. We only do that in test files.  The room file also does not need to be required in our `House` class file, because we don't need to instantiate a `Room` object in this file. Nothing in the interaction pattern asks us to do that, BUT we can still pass in `Room` objects that we create in our test via a method we create in our `House` file.

test/house_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'minitest'
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/room'
+require './lib/house'
+
+class HouseTest < Minitest::Test
+
+  def test_it_exists
+    house = House.new('$400000', '123 sugar lane', [])
```
timomitchel, line 20: This is a solid test.

timomitchel (commented):
I highly recommend that you spend time working on this challenge with a mentor or an upper mod student. Then, I would completely delete the challenge an attempt it on your own. Hours of practice is the only way to progress. This is well below our expectations for this challenge. But, with time and effort we know you can improve!

PR #27, opened 2020-02-04
Zach Holcomb independent challenge

zachholcomb:
Almost finished iteration 3, not sure how to create hash of arrays.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,7 @@
+class House
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,5 +1,5 @@
 class House
-attr_reader :price
+attr_reader :price, :address
   def initialize(price, address)
     @price = price
     @address = address
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -11,11 +11,18 @@ attr_reader :price, :address, :rooms
   end
 
   def above_market?
-    if @price.to_i > 500000
+    if @price.delete('$').to_i > 500000
       true
     else
       false
     end
   end
 
+  def price_per_square_foot
+    house_total_square_feet = 0
+    @rooms.each do |room|
+      house_total_square_feet += room.area
+    end
+    ((@price.delete('$').to_i / house_total_square_feet.to_f)).round(2)
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -9,4 +9,13 @@ attr_reader :price, :address, :rooms
   def add_room(room)
     @rooms << room
   end
+
+  def above_market?
+    if @price.to_i > 500000
+      true
+    else
+      false
+    end
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -25,4 +25,11 @@ attr_reader :price, :address, :rooms
     end
     ((@price.delete('$').to_i / house_total_square_feet.to_f)).round(2)
   end
+
+  def rooms_sorted_by_area
+    @rooms.sort! do |room|
+      room.area
+    end
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -32,4 +32,10 @@ attr_reader :price, :address, :rooms
     end
   end
 
+  def rooms_by_category
+    rooms_by_category = {}
+    @rooms.each do |room|
+      rooms_by_category[room.category] = [room]
+    end
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,5 +1,5 @@
 class House
-
+attr_reader :price
   def initialize(price, address)
     @price = price
     @address = address
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -5,4 +5,8 @@ attr_reader :price, :address, :rooms
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
+    @rooms << room
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,7 +1,8 @@
 class House
-attr_reader :price, :address
+attr_reader :price, :address, :rooms
   def initialize(price, address)
     @price = price
     @address = address
+    @rooms = []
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,41 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market?
+    if @price.delete('$').to_i > 500000
```
timomitchel, line 14: This is a solid setup for this method, but I would probably do my conversion to an integer right in initialize, so that my instance variable henceforth is the data type I can perform math on. Another way to right this that is a bit cleaner would be:
```
return true if @price.delete('$').to_i > 500000
false
```

lib/house.rb
```diff
@@ -0,0 +1,41 @@
...
+      false
+    end
+  end
+
+  def price_per_square_foot
+    house_total_square_feet = 0
+    @rooms.each do |room|
+      house_total_square_feet += room.area
+    end
+    ((@price.delete('$').to_i / house_total_square_feet.to_f)).round(2)
```
timomitchel, line 26: This shows how if you do the `@price.delete('$').to_i` in `initialize` you can avoid duplicating work. That would make this code more DRY (Don't repeat yourself).

lib/house.rb
```diff
@@ -0,0 +1,41 @@
...
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market?
```
timomitchel, line 13: Another thing to remember: Make sure to name the methods exactly what they are in the spec. This way they pass our tests if they are written correctly.

lib/house.rb
```diff
@@ -0,0 +1,41 @@
...
+  def price_per_square_foot
+    house_total_square_feet = 0
+    @rooms.each do |room|
+      house_total_square_feet += room.area
+    end
+    ((@price.delete('$').to_i / house_total_square_feet.to_f)).round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort! do |room|
```
timomitchel, line 30: We wouldn't want to use the destructive `!` operator here. This will change the order of the `@rooms` objects array to be sorted in descending order by area and now be stored that way moving forward. We don't want to change what our `@rooms` instance variable is holding, we just want this method to do the work to our array and return that value, but not change how our room objects are stored in our instance variable. This could cause unexpected behavior later on.

lib/house.rb
```diff
@@ -0,0 +1,41 @@
...
+      room.area
+    end
+  end
+
+  def rooms_by_category
+    rooms_by_category = {}
+    @rooms.each do |room|
+      rooms_by_category[room.category] = [room]
```
timomitchel, line 38: This is super close, but we need a way to add to our array if the key has already been created via our iteration. This will accomplish that: 
```
rooms_by_category = {}
@rooms.each do |room|
if rooms_by_category[room.category]
rooms_by_category[room.category] << room  
else  
rooms_by_category[room.category] = [room]    
end
end
```

lib/room.rb
```diff
@@ -1,12 +1,25 @@
...
+    @is_painted = false
   end
 
   def area
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    if @is_painted == false
```
timomitchel, line 15: Once again, a guard statement will make this a bit cleaner:
```
return true if @is_painted
false
```

test/house_test.rb
```diff
@@ -0,0 +1,80 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/house'
+require './lib/room'
+
+class HouseTest < Minitest::Test
+  def setup
+    @house = House.new("$400000", "123 sugar lane")
```
timomitchel, line 8: Great use of the setup method here.

test/house_test.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def test_it_exists
+
+    assert_instance_of House, @house
+  end
+
+  def test_it_has_a_price
+
+    assert_equal "$400000", @house.price
```
timomitchel, line 22: Pay close attention to the spec - this should have been an `integer`. 
```
pry(main)> house.price
#=> 400000
```

test/house_test.rb
```diff
@@ -0,0 +1,80 @@
...
+  def test_it_has_rooms
+
+    assert_equal [], @house.rooms
+  end
+
+  def test_it_can_add_rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+
+    assert_equal [@room_1, @room_2], @house.rooms
```
timomitchel, line 39: Nice work here.

test/house_test.rb
```diff
@@ -0,0 +1,80 @@
...
+    assert_equal 210.53, @house.price_per_square_foot
+  end
+
+  def test_it_can_sort_rooms_by_area
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+
+    assert_equal [@room_4, @room_3, @room_2, @room_1], @house.rooms_sorted_by_area
```
timomitchel, line 64: Clean and effective.

test/house_test.rb
```diff
@@ -0,0 +1,80 @@
...
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+    rooms_by_category = {
+      :bedroom => [@room_1, @room_2],
+      :living_room => [@room_3],
+      :basement => [@room_4]}
+
+    assert_equal  rooms_by_category, @house.rooms_by_category
```
timomitchel, line 77: Nice setup here!

timomitchel (commented):
Overall a solid job on this challenge Zach. Your score was lower than your understanding because of how you named some of your methods. Make sure to name exactly how they are in the spec, and try not to move on from one iteration until you've completed all the methods the spec calls for. Looks like you skipped over the `details` method.

PR #28, opened 2020-02-04
Kelsha Darby

kelshadarby:
I was not able to get my details test to pass but it was passing in pry using the hash and the details method. I'm not sure if assert_equal was the wrong choice for this one.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,44 @@
+require './lib/room'
+
+class House
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+end
+
+
+
+# pry(main)> require './lib/room'
+# #=> true
+#
+# pry(main)> require './lib/house'
+# #=> true
+#
+# pry(main)> house = House.new("$400000", "123 sugar lane")
+# #=> #<House:0x00007fccd30375f8...>
+#
+# pry(main)> house.price
+# #=> 400000
+#
+# pry(main)> house.address
+# #=> "123 sugar lane"
+#
+# pry(main)> house.rooms
+# #=> []
+#
+# pry(main)> room_1 = Room.new(:bedroom, 10, '13')
+# #=> #<Room:0x00007fccd29b5720...>
+#
+# pry(main)> room_2 = Room.new(:bedroom, 11, '15')
+# #=> #<Room:0x00007fccd2985f48...>
+#
+# pry(main)> house.add_room(room_1)
+#
+# pry(main)> house.add_room(room_2)
+#
+# pry(main)> house.rooms
+# #=> [#<Room:0x00007fccd29b5720...>, #<Room:0x00007fccd2985f48...>]
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -11,4 +11,55 @@ class House
   def add_room(room)
     @rooms << room
   end
+
+  def above_market_average?
+    if @price > 500000
+      true
+    else
+      false
+    end
+  end
+
+  def area
+    room_areas = []
+    @rooms.each do |room|
+      room_areas << room.area
+    end
+    room_areas.sum
+  end
+
 end
+
+
+
+# pry(main)> room_1 = Room.new(:bedroom, 10, '13')
+# #=> #<Room:0x00007fccd29b5720...>
+#
+# pry(main)> room_2 = Room.new(:bedroom, 11, '15')
+# #=> #<Room:0x00007fccd2985f48...>
+#
+# pry(main)> room_3 = Room.new(:living_room, 25, '15')
+# #=> #<Room:0x00007fccd383c2d0...>
+#
+# pry(main)> room_4 = Room.new(:basement, 30, '41')
+# #=> #<Room:0x00007fccd297dc30...>
+#
+# pry(main)> house.add_room(room_1)
+#
+# pry(main)> house.add_room(room_2)
+#
+# pry(main)> house.add_room(room_3)
+#
+# pry(main)> house.add_room(room_4)
+#
+# pry(main)> house.rooms_from_category(:bedroom)
+# #=> [#<Room:0x00007fccd29b5720...>, #<Room:0x00007fccd2985f48...>]
+#
+# pry(main)> house.rooms_from_category(:basement)
+# #=> [#<Room:0x00007fccd297dc30...>]
+#
+# pry(main)> house.area
+# #=> 1900
+#
+# pry(main)> house.details
+# #=> {"price" => 400000, "address" => "123 sugar lane"}
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -28,38 +28,15 @@ class House
     room_areas.sum
   end
 
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
+      if room.category == category
+        room
+      end
+    end
+  end
+
+  def details
+    details = {"price" => @price, "address" => @address}
+  end
 end
-
-
-
-# pry(main)> room_1 = Room.new(:bedroom, 10, '13')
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,42 @@
+class House
+
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price[1..-1].to_i
```
aziobrow, line 6: nice!

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    if @price > 500000
+      true
+    else
+      false
+    end
```
aziobrow, line 20: This is great.

It could also be written in a single line as:

`@price > 500000`, since that statement in and of itself returns a boolean value

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def area
+    room_areas = []
+    @rooms.each do |room|
+      room_areas << room.area
+    end
+    room_areas.sum
```
aziobrow, line 28: This is great, and works well.  `Sum` is an enumerable of its own, and could have been used instead of `each. 

```
def area 
    @rooms.sum do |room|
         room.area
     end
end
```

it would iterate over each room and sum the area of each room to get you the total

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
+      if room.category == category
+        room
+      end
+    end
```
aziobrow, line 36: Nice job choosing the best tool for the job! `find_all` is a great enumerable here. `find_all` only takes a boolean condition though, so the way this should actually  be written is as follows:

```
   @rooms.find_all do |room|
         room.category == category 
    end
end 
```

I think in this case, you got a little bit lucky, and it works anyway, but you can get yourself into trouble, setting yourself up for false positives, by putting anything other than a conditional into a `find_all`

lib/house.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def details
+    details = {"price" => @price, "address" => @address}
```
aziobrow, line 40: 👍

lib/room.rb
```diff
@@ -1,12 +1,27 @@
...
+class Room
+
+  attr_reader :category
 
   def initialize(category, length, width)
     @category = category
-    @length = 10
+    @length = length
     @width = width
+    @paint = nil
```
aziobrow, line 9: we'd probably set this to false (i.e., when a house is very first created, it is unpainted).  This is logical, and additionally, it makes it a little easier since you don't need the conditional in your `is_painted?` method.

lib/room.rb
```diff
@@ -1,12 +1,27 @@
...
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    if @paint == nil
+      false
+    else
+      true
+    end
```
aziobrow, line 21: If you default `@painted` to false, this whole method can just be:

```
def is_painted?
    @painted 
end
```

lib/room.rb
```diff
@@ -1,12 +1,27 @@
...
+  end
+
+  def paint
+    @paint = true
```
aziobrow, line 25: 👍

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+require './lib/room'
+require './lib/house'
+
+class HouseTest < Minitest::Test
+
+  def test_it_exists_and_has_attributes
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_instance_of House, house
+    assert_equal 400000, house.price
```
aziobrow, line 12: nice catch here

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def test_it_can_have_rooms
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+
+    assert_instance_of Room, room_1
+
+    room_2 = Room.new(:bedroom, 11, '15')
+
+    assert_instance_of Room, room_2
```
aziobrow, line 31: It doesn't hurt to do two different rooms, but unless there's something substantially different about these two rooms, it probably doesn't do you any good, either, so we can just test one.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+
+    assert_equal [room_1, room_2], house.rooms
+
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2, room_3, room_4], house.rooms
```
aziobrow, line 48: Same here-- I like the thoroughness of your thought process here, but unless there's something different about adding 4 rooms instead of 2, there's probably no reason to do the extra two rooms/assertions here.

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal [room_1, room_2], house.rooms_from_category(:bedroom)
+    assert_equal [room_4], house.rooms_from_category(:basement)
+    assert_equal [room_3], house.rooms_from_category(:living_room)
```
aziobrow, line 64: doing more than one category here is a good idea though! does it work if you pass in a category that doesn't exist? (it should, but that could be a good thing to test! although for a timed exercise, i wouldn't worry about edge case testing too much)

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    room_3 = Room.new(:living_room, 25, '15')
+    room_4 = Room.new(:basement, 30, '41')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    house.add_room(room_3)
+    house.add_room(room_4)
+
+    assert_equal 1900, house.area
```
aziobrow, line 78: 👍

test/house_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def test_it_can_be_above_market_average?
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house.above_market_average?
```
aziobrow, line 84: in this case, probably a good idea to create another house that is above market average and make sure that the other possibility works as well.

test/room_test.rb
```diff
@@ -9,16 +9,24 @@ def test_it_exists
...
   end
+
+  def test_it_can_tell_if_room_can_be_painted
+    room = Room.new(:bedroom, 10, '13')
+
+    assert_equal false, room.is_painted?
+
+    room.paint
+
+    assert_equal true, room.is_painted?
```
aziobrow, line 30: nicely done-- good progression here

aziobrow (commented):
Great work! You got through iteration 3 accurately and in time, which is exactly what we're hoping for.  

Looks like you didn't get to the three methods in iteration 4, which is completely fine.  Loved your attention to detail-- you did a great job of both implementation and following the interaction pattern exactly.  I left some inline comments with a bit more detail. 

lib/house.rb
kelshadarby: Woahhh! I didn't know sum was also an enumerable! Thanks for calling that out!

PR #29, opened 2020-02-04
Alex Latham

alex-latham:


lib/house.rb
```diff
@@ -0,0 +1,46 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
```
aziobrow, line 5: be careful-- the interaction pattern wants price to be an integer, and yours is still the string.  Be sure your tests/implementation match the interaction pattern exactly.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    @price[1..-1].to_i > 500000 ? true : false
```
aziobrow, line 15: This is nice, although technically, you don't need the ternary here. 

if you just did:
```
def above_market_average?
     @price[1..-1].to_i > 500000
end
```

it would return the correct boolean all on its own, without you explicitly needing to return true or false

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.select {|room| room.category == category}
```
aziobrow, line 19: great enumerable for this use case

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def area
+    @rooms.sum {|room| room.area}
```
aziobrow, line 23: perfect

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def details
+    {"price" => @price[1..-1].to_i, "address" => @address}
```
aziobrow, line 27: 👍

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def price_per_square_foot
+    (@price[1..-1].to_f / area).round(2)
```
aziobrow, line 31: nice work, and good job using the return value of another method here.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def rooms_sorted_by_area
+    (@rooms.sort_by {|room| room.area}).reverse
```
aziobrow, line 35: Great. no need for the parentheses here, since that's all part of the same method call.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def rooms_by_category
+    rmc = {}
```
aziobrow, line 39: this is not a good variable name. coding the ruby way is to be explicit and readable 🙂

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+    categories = @rooms.uniq {|room| room.category}
```
aziobrow, line 40: the way you have it, this is returning to you an array of rooms, not categories.  Using `pry` here would help you check your assumptions. 

`categories = @rooms.map {|room| room.category}.uniq` would have given you what you wanted.

lib/house.rb
```diff
@@ -0,0 +1,46 @@
...
+    categories.each do |category|
+      rmc[category] = rooms_from_category(category)
```
aziobrow, line 42: nice job setting key/value pairs, and nice job using the return value of another method here. Ultimately, this doesn't work, because you're passing in a room instead of a category here, which means that every value is an empty array (since it can't find any rooms with a category that matches a room object), but if `categories` had been correct, this would have been perfect.

lib/room.rb
```diff
@@ -1,12 +1,22 @@
...
-    @length * @width
+    @length * @width.to_i
+  end
+
+  def is_painted?
+    @painted
+  end
+
+  def paint
+    @painted = true
```
aziobrow, line 20: nice job with this test

test/house_test.rb
```diff
@@ -0,0 +1,134 @@
...
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_instance_of House, house
+  end
+
+  def test_it_has_a_price
+    house = House.new("$400000", "123 sugar lane")
+
+    assert_equal "$400000", house.price
```
aziobrow, line 16: be careful that you match the interaction pattern exactly. this should have been an integer.

test/house_test.rb
```diff
@@ -0,0 +1,134 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+
+    assert_equal [], house.rooms
+
+    house.add_room(room_1)
+    house.add_room(room_2)
+
+    assert_equal [room_1, room_2], house.rooms
```
aziobrow, line 35: nice job with the progression testing here

test/house_test.rb
```diff
@@ -0,0 +1,134 @@
...
+  end
+
+  def test_it_can_tell_whether_house_is_above_market_average
+    house1 = House.new("$400000", "123 sugar lane")
+
+    assert_equal false, house1.above_market_average?
+
+    house2 = House.new("$600000", "123 sugar lane")
+    assert_equal true, house2.above_market_average?
```
aziobrow, line 44: nice job testing both options here

test/room_test.rb
```diff
@@ -9,16 +9,24 @@ def test_it_exists
...
+
+  def test_can_test_if_room_is_painted
+    room = Room.new(:kitchen, 10, '9')
+
+    assert_equal false, room.is_painted?
+
+    room.paint
+
+    assert_equal true, room.is_painted?
+  end
```
aziobrow, line 31: 👍

aziobrow (commented):
Very nice work on this! The two that you didn't get passing were `price` (because you didn't follow the interaction pattern precisely) and the last method that it looks like you were working on when time was up (`rooms_by_category`).  You did get completely and mostly accurately through Iteration 3, though, so your progress is exactly where we'd hope for it to be.   I left you some inline comments with more detail. Let me know if you have any questions!

PR #30, opened 2020-02-04
Joshua Tukman 

Joshua-Tukman:
SO FRUSTRATED! 
I worked for 45 minutes to create test for 'returning array with rooms in category'. Seemed so simple but couldn't figure it out.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,9 +1,10 @@
 class House
-  attr_reader :price, :address
+  attr_reader :price, :address, :rooms  
 
   def initialize(price, address)
     @price = price
     @address = address
+    @rooms = []
   end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -11,5 +11,15 @@ class House
     @rooms << room
   end
 
+  def above_market_average?
+    if @price.delete!("$").to_i > 500000
+      true
+    else
+      false
+    end
+  end
+
+
+
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,9 @@
+class House
+  attr_reader :price, :address
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -19,7 +19,13 @@ class House
     end
   end
 
-
-
-
+  def rooms_from_category(category)
+    rooms_in_category = []
+    @rooms.each do |room|
+      if room.category == category
+        rooms_in_category << room
+      end
+    end
+    rooms_in_category
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,10 +1,15 @@
 class House
-  attr_reader :price, :address, :rooms  
+  attr_reader :price, :address, :rooms
 
   def initialize(price, address)
     @price = price
     @address = address
     @rooms = []
   end
+
+  def add_room(room)
+    @rooms << room
+  end
+
 
 end
--- a/test/house_test.rb
+++ b/test/house_test.rb
@@ -31,20 +31,47 @@ class HouseTest < Minitest::Test
     assert_equal [@room_1], @house.add_room(@room_1)
     @room_2 = Room.new(:bedroom, 11, '15')
     assert_equal [@room_1, @room_2], @house.add_room(@room_2)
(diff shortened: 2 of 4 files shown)


test/house_test.rb
```diff
@@ -0,0 +1,70 @@
...
+  def setup
+    @house = House.new("$400000", "123 sugar lane")
+  end
+
+  def test_it_exists
+    assert_instance_of House, @house
+  end
+
+  def test_it_has_a_price
+    assert_equal "$400000", @house.price
```
timomitchel, line 18: Pay close attention to the spec. This should have been converted to an integer at this point: 

```
pry(main)> house.price
#=> 400000
```

test/house_test.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def test_it_has_an_address
+    assert_equal "123 sugar lane", @house.address
+  end
+
+  def test_it_initializes_with_empty_rooms_array
+    assert_equal [], @house.rooms
+  end
+
+  def test_it_can_add_rooms_and_store_in_rooms
```
timomitchel, line 29: Nice job testing this method thoroughly here. We actually don't need the test on line 31 or 33, because that's a command method `.add_room` that we use to setup our query method `.rooms` that our command method is changing. A better setup would be: 
```
    room_1 = Room.new(:bedroom, 10, '13')
    room_2 = Room.new(:bedroom, 11, '15')
    assert_equal [],  @house.rooms
    @house.add_room(room_1)
    assert_equal [room_1], @house.rooms
    @house.add_room(room_2)
    assert_equal [room_1, room_2], @house.rooms
```

test/house_test.rb
```diff
@@ -0,0 +1,70 @@
...
+    assert_equal [@room_1, @room_2], @house.add_room(@room_2)
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_it_is_above_market_average
+    assert_equal false, @house.above_market_average?
+  end
+
+  def test_can_return_rooms_in_category
+    @room_1 = Room.new(:bedroom, 10, '13')
```
timomitchel, line 42: We don't want to create instance variables in a single test. If you want to create instance variables that the entire test class has access to, then do it in the `setup` method. Otherwise create local variables that are scoped to specific test methods.

test/house_test.rb
```diff
@@ -0,0 +1,70 @@
...
+    @room_2 = Room.new(:bedroom, 11, '15')
+    @room_3 = Room.new(:living_room, 25, '15')
+    @room_4 = Room.new(:basement, 30, '41')
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+    assert_equal [@room_1, @room_2], @house.rooms_from_category(:bedrooms)
```
timomitchel, line 50: This is a great test (other than the improper use of the `@room_1, @room_2, etc` instance variables.

test/room_test.rb
```diff
@@ -3,22 +3,32 @@
...
+    assert_equal 130, @room.area
+    assert_equal 180, @room2.area
+  end
+
+  def test_it_is_painted
+    assert_equal false, @room.is_painted?
+  end
+
+  def test_it_can_be_painted
+    @room.paint
```
timomitchel, line 30: Nice job with this.

lib/house.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def above_market_average?
+    if @price.delete!("$").to_i > 500000
+      true
+    else
+      false
+    end
+  end
+
+  def rooms_from_category(category)
```
timomitchel, line 22: Great job using each here to get the job done. `find_all` would save you a few lines here: 
```
@rooms.find_all do |room|
room.category == category
end
```

timomitchel (commented):
Overall a really solid job on this challenge Josh. I know it can be frustrating to be stuck on one thing for 45 minutes. In these times try to reach for your debugging strategies and experiment within pry! And, don't be afraid to take 5 and then come back to the problem. This usually helps you view it with a fresh perspective!

test/house_test.rb
timomitchel: Ah I see your issue here, You're passing `:bedrooms` as an argument instead of `:bedroom`. This is where copy and pasting directly from the interaction pattern can help you avoid silly mistakes that cause you to get stuck!

PR #31, opened 2020-02-04
Steven Meyers

steve-meyers:
Tried my hardest!

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,13 @@
+class House
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+
+
+
+
+
+
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -23,6 +23,16 @@ def above_market_average?
   end
 end
 
+def rooms_from_category(category)
+  rooms_in_category = []
+  @rooms.each do |room|
+    if category == room.category
+      rooms_in_category << room
+    end
+  end
+    rooms_in_category
+end
+
 
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,15 +1,19 @@
 class House
+  attr_reader :rooms
 
   def initialize(price, address)
     @price = price
     @address = address
+    @rooms = []
   end
 
   def price
     @price.to_i
   end
 
-
+def add_room(room)
+  @rooms << room
+end
 
 
 
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -15,6 +15,14 @@ def add_room(room)
   @rooms << room
 end
 
+def above_market_average?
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,38 @@
+class House
+  attr_reader :rooms
```
memcmahon, line 2: Keep a close eye on the interaction pattern - we also wanted an attr_reader for `address`

lib/house.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def price
+    @price.to_i
```
memcmahon, line 11: So close, but if you call `.to_i` on a string that begins with a non-integer character, it will return `0`.  It would be better to remove that character and then `to_i`.  This would be one option: `@price[1..-1].to_i`

lib/house.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+def add_room(room)
+  @rooms << room
+end
+
+def above_market_average?
```
memcmahon, line 18: the logic here is correct, but this method will always return false, because `@price.to_i` returns `0` (see my note above)

lib/house.rb
```diff
@@ -0,0 +1,38 @@
...
+  if @price.to_i > 500000
+    true
+  else
+    false
+  end
+end
+
+def rooms_from_category(category)
```
memcmahon, line 26: This method looks GREAT!

lib/room.rb
```diff
@@ -1,12 +1,26 @@
-class Rom
+class Room
```
memcmahon, line 1: Great job with this class - especially the methods `is_painted?` and `paint`

test/house_test.rb
```diff
@@ -0,0 +1,55 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/room'
+require './lib/house'
+require 'pry'
+
+class HouseTest < Minitest::Test
+  def test_it_exists
+    house = House.new("400000", "123 sugar lane")
```
memcmahon, line 9: Keep a super close eye on the interaction pattern - a house is created with a price of `"$XXXXX"`.  Including the `$` would have shown you that some of the methods that are passing your tests do not correctly model the behavior outlined in the interaction pattern.  Creating your house like this is giving you a false positive on the price, and above_market_average? tests

memcmahon (commented):
Hey Steve - there is some great stuff here, but not quite where we would like you to be.  Make sure to keep a close eye on the interaction pattern next time, and I would encourage you to work with your mento on cleaning this up, and finishing it :)

PR #33, opened 2020-02-04
Stephanie Friend

StephanieFriend:


lib/house.rb
```diff
@@ -0,0 +1,54 @@
+class House
+
+  attr_reader :address, :price, :rooms
+
+  def initialize(price, address)
+    @price = price[1..-1].to_i
```
aziobrow, line 6: nice

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    rooms << room
+  end
+
+  def above_market_average?
+    price > 500000
```
aziobrow, line 16: nice!

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def rooms_from_category(category)
+    rooms.find_all { |room| room.category == category }
```
aziobrow, line 20: perfect enumerable for this job

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def area
+    total_area = []
+    rooms.each { |room| total_area << room.area }
+    total_area.sum
```
aziobrow, line 26: This works fine, but remember-- sum is an enumerable as well, so you're iterating over the whole collection twice (not a big deal when you only have 4 rooms, but if you have 4 thousand or 4 million, it would be a real big deal).  So instead of iterating twice here, you could just do:

```
def area 
       rooms.sum { |room| total_area << room.area }
end
```

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def details
+    Hash.new
```
aziobrow, line 30: you're not actually using this Hash.new here-- you're not saving it to a variable or doing anything else with it, so you can just take it out.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+    {"price" => price, "address" => address}
+  end
+
+  def price_per_square_foot
+    (price.to_f / area.to_f).round(2)
```
aziobrow, line 35: nice job using the return value of another method here.  Tiny point, but you only need one of those values to be a float in order to get a float in your answer, so you only need to do `to_f` on either price or area, but not both.

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def rooms_sorted_by_area
+    own_area = {}
+    rooms.each do |room|
+      own_area[room] = room.area
+    end
+    own_area.keys.sort_by { |value| own_area[value] }.reverse
```
aziobrow, line 43: This was a fairly complex solution, which you got working! Impressive!

Remember that you can sort a collection of objects by any attribute, so I'd do this like this:

```
def rooms_sorted_by_area
   rooms.sort_by {|room| room.area}.reverse
end
```

lib/house.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  #I keep getting basement first?
+  def rooms_by_category
+    sorted_rooms = {}
+    rooms.each do |room|
+      sorted_rooms[room.category] = [room]
```
aziobrow, line 50: you're setting the room category as the key equal to an array always with one element inside. So if there are multiple rooms for that category, you're always going to get the last one, because that array can only ever have one room, based on the way you wrote it. 

You could either check out `group_by`, which is probably how I would have done it (http://gregpark.io/blog/ruby-group-by/)

Or if you do it this way, you could have done something like this:

```
def rooms_by_category
    sorted_rooms = {}
    rooms.each do |room|
      sorted_rooms[room.category] = rooms_from_category(room.category)
    end
end
```
This isn't as efficient as group_by, but I'm fairly certain it would work!

test/house_test.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    assert_equal [room_1, room_2], house.rooms
+  end
+
+  def test_it_can_return_if_house_is_above_market_average
+    house_1 = House.new("$400000", "123 sugar lane")
+    house_2 = House.new("$500500", "124 sugar lane")
+
+    assert_equal false, house_1.above_market_average?
+    assert_equal true, house_2.above_market_average?
```
aziobrow, line 43: nice job testing both paths here

test/house_test.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    house.add_room(room_4)
+    house.add_room(room_1)
+    house.add_room(room_3)
+    house.add_room(room_2)
+
+    expected = {:bedroom => [room_1, room_2], :living_room => [room_3], :basement => [room_4]}
+
+    assert_equal expected, house.rooms_by_category
+  end
```
aziobrow, line 134: nice job with these tests

test/room_test.rb
```diff
@@ -9,16 +9,28 @@ def test_it_exists
...
+
+    assert_equal false, room.is_painted?
+  end
+
+  def test_room_can_be_painted
+    room = Room.new(:bedroom, 10, '13')
+
+    room.paint
+
+    assert_equal true, room.is_painted?
```
aziobrow, line 34: great!

aziobrow (commented):
Great job, Stephanie! The only method you didn't quite get working was the `rooms_by_category`.  Things look really good-- you did a nice job of both implementation and testing.  I left you some inline comments with specific details/suggestions. 

let me know if you have any questions!

PR #34, opened 2020-02-04
Ryan Camp

cmpprg:


lib/house.rb
```diff
@@ -0,0 +1,58 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(value, address)
+    @price = (value.sub("$", "")).to_i
```
timomitchel, line 5: Nice job putting this conversion in initialize. This helps to avoid duplicating the work later on.

lib/house.rb
```diff
@@ -0,0 +1,58 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    return @price > 500000
```
timomitchel, line 15: Nice job, but no need for the explicit `return` here.

lib/house.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all do |room|
```
timomitchel, line 19: Great job choosing the right enum here.

lib/house.rb
```diff
@@ -0,0 +1,58 @@
...
+      room.category == category
+    end
+  end
+
+  def area
+    total_area = 0
```
timomitchel, line 25: Nice use of each in this method. You could also pass `sum` a block: 
`rooms.sum {|room| room.area}` to make this even more readable.

lib/house.rb
```diff
@@ -0,0 +1,58 @@
...
+    rooms.each do |room|
+      total_area += room.area
+    end
+    total_area
+  end
+
+  def details
+    {"price"=>@price, "address"=>@address}
```
timomitchel, line 33: Clean and effective.

lib/house.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def price_per_square_foot
+    (@price / area.to_f).round(2)
+  end
+
+  def rooms_sorted_by_area
+    (rooms.sort_by {|room| room.area}).reverse
```
timomitchel, line 41: Nice job here, but no need for the parentheses. You could use `sort` to avoid the `reverse`.

lib/house.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def rooms_by_category
+    list_of_categories = []
```
timomitchel, line 45: Nice job coming up with a solution here. Check out the enum/method `group_by` to save you some work!

test/house_test.rb
```diff
@@ -0,0 +1,128 @@
+require "minitest/autorun"
+require "minitest/pride"
+require "./lib/house.rb"
+require "./lib/room.rb"
+
+class HouseTest < Minitest::Test
+  def test_it_exists
```
timomitchel, line 7: You should consider using a `setup` method with instance variables to create your objects for this test so that you can avoid duplicating work and DRY this up a bit. It will also save you precious time in these challenges!

test/house_test.rb
```diff
@@ -0,0 +1,128 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    room1 = Room.new(:bedroom, 10, '13')
+    room2 = Room.new(:bedroom, 11, '15')
+    house.add_room(room1)
+
+    assert_equal [room1], house.rooms
+
+    house.add_room(room2)
+
+    assert_equal [room1, room2], house.rooms
```
timomitchel, line 31: Nice and robust pattern here!

test/house_test.rb
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  def test_house_above_market_average_method
+    house = House.new("$400000", "123 sugar lane")
+    house2 = House.new("$600000", "125 sugar lane")
+
+    assert_equal false, house.above_market_average?
+    assert_equal true, house2.above_market_average?
```
timomitchel, line 39: Great job testing both scenarios here. Another test that passes the you the exact value that you are comparing, in this case `$500000` would be a nice addition.

test/house_test.rb
```diff
@@ -0,0 +1,128 @@
...
+    room1 = Room.new(:bedroom, 10, '13')
+    room2 = Room.new(:bedroom, 11, '15')
+    room3 = Room.new(:living_room, 25, '15')
+    room4 = Room.new(:basement, 30, '41')
+    house.add_room(room1)
+    house.add_room(room2)
+    house.add_room(room3)
+    house.add_room(room4)
+
+    assert_equal [room1, room2], house.rooms_from_category(:bedroom)
```
timomitchel, line 53: Nice job testing multiple categories here.

test/house_test.rb
```diff
@@ -0,0 +1,128 @@
...
+    house = House.new("$400000", "123 sugar lane")
+    room1 = Room.new(:bedroom, 10, '13')
+    room2 = Room.new(:bedroom, 11, '15')
+    room3 = Room.new(:living_room, 25, '15')
+    room4 = Room.new(:basement, 30, '41')
+    house.add_room(room1)
+    house.add_room(room2)
+    house.add_room(room3)
+    house.add_room(room4)
+    details = {"price"=>400000, "address"=>"123 sugar lane"}
```
timomitchel, line 81: This works, but the convention is to name this variable `expected` when we set up a value in our test like this.

timomitchel (commented):
Great job on this challenge Ryan! You're looking strong. Make sure to help out your classmates that aren't as comfortable yet. Nothing solidifies our mastery like teaching and putting code into words!

PR #35, opened 2020-02-04
Jesse Gietzen

elguapogordo:
complete through iteration 3.
100% functionality on each test.
iteration 4 not started.

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,2 @@
+class House
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,10 +1,11 @@
 class House
 
-  attr_reader :price, :address
+  attr_reader :price, :address, :rooms
 
   def initialize(price, address)
     @price = price
     @address = address
+    @rooms = []
   end
 
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -8,4 +8,8 @@ class House
     @rooms = []
   end
 
+  def add_room(room)
+    @rooms << room
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,6 +1,6 @@
 class House
 
-  attr_reader :price
+  attr_reader :price, :address
 
   def initialize(price, address)
     @price = price
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,2 +1,8 @@
 class House
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+  end
+  
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -19,4 +19,9 @@ class House
     house_price.join.to_i > 500000
   end
 
+  def rooms_from_category(category)
+    require 'pry'; binding.pry
+    rooms_from_category = @rooms.find_all { |room| room.category == category }
+  end
+
 end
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,36 @@
+class House
+
+  attr_reader :price, :address, :rooms
```
memcmahon, line 3: Keep an eye on the interaction pattern - the price should have been a separate method that returned price as an integer (`400000` instead of `"$4000000"`)

lib/house.rb
```diff
@@ -0,0 +1,36 @@
...
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def above_market_average?
+    house_price = @price.chars.map { |char| char.to_i }
```
memcmahon, line 16: great job getting the house_price in integers - since you use it multiple times though, it really deserves to be a separate method (see my note on your attr_readers)

lib/house.rb
```diff
@@ -0,0 +1,36 @@
...
+    house_price.join.to_i > 500000
+  end
+
+  def rooms_from_category(category)
+    rooms_from_category = @rooms.find_all { |room| room.category == category }
```
memcmahon, line 21: Great use of find_all, but you don't really need to assign the result to a variable - you can let the return value of the find_all be the return value of the method.

lib/house.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def area
+    @rooms.map { |room| room.area }.sum
```
memcmahon, line 25: Great combination of `map` and `sum`.  You _could_ also use sum like this:

```ruby
@rooms.sum { |room| room.area }
```

lib/room.rb
```diff
@@ -1,12 +1,22 @@
-class Rom
+class Room
+
+  attr_reader :category
+  attr_accessor :is_painted
+  alias_method :is_painted?, :is_painted
```
memcmahon, line 5: great use of an alias_method!  Another way to accomplish this would be to define a method explicitly:

```ruby
def is_painted?
  @is_painted
end
```

test/house_test.rb
```diff
@@ -0,0 +1,75 @@
...
+class HouseTest < Minitest::Test
+
+  def test_it_exists
+    house = House.new("$400000", "123 sugar lane")
+    assert_instance_of House, house
+  end
+
+  def test_it_has_price
+    house = House.new("$400000", "123 sugar lane")
+    assert_equal house.price, "$400000"
```
memcmahon, line 15: Again, keep an eye on the interaction pattern for the expected return value of specific methods!

test/house_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    assert_equal house.rooms, []
+  end
+
+  def test_can_we_add_rooms
+    house = House.new("$400000", "123 sugar lane")
+    room_1 = Room.new(:bedroom, 10, '13')
+    room_2 = Room.new(:bedroom, 11, '15')
+    house.add_room(room_1)
+    house.add_room(room_2)
+    assert_equal room_1, house.rooms[0]
```
memcmahon, line 34: You could also combine these two assertions like this: 
```ruby
assert_equal [room_1, room_2], house.rooms
```

memcmahon (commented):
Hey Jesse - this looks really good!  Take a look at my notes, but overall, it looks like you are on track for where we would expect you to be 🙂

PR #37, opened 2020-02-04
Tyler Tomlinson

tylertomlinson:
Finished Iteration 4

### DIFF
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -44,13 +44,4 @@ class House
   def rooms_sorted_by_category
     @rooms.group_by {|room| room.category}
   end
-
-
-
-
-
-
-
-
-
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -1,0 +1,13 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+  end
+
+  def add_room(room)
+    @rooms << room
+  end 
+end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -20,4 +20,7 @@ class House
     @price > @above_market_price
   end
 
+  def rooms_from_category(category)
+    @rooms.find_all {|room| room.category == category}
+  end
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -2,10 +2,12 @@ class House
   attr_reader :price, :address, :rooms
 
   def initialize(price, address)
+    @attributes = ()
     @price = price
     @address = address
     @rooms = []
     @above_market_price = "$500,000"
+    @details = Hash.new
   end
 
   def add_room(room)
@@ -23,4 +25,19 @@ class House
   def rooms_from_category(category)
     @rooms.find_all {|room| room.category == category}
   end
+
+  def area
+    @rooms.sum {|room| room.area}
+  end
+
+  def details(price, address)
+    key_values = ["price", price ], ["address", address]
+    key_values.to_h
+  end
+
+  
+
+
+
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -2,18 +2,18 @@ class House
   attr_reader :price, :address, :rooms
 
   def initialize(price, address)
-    @price = price
+    @price = reformated_price(price)
     @address = address
     @rooms = []
-    @above_market_price = "$500,000"
+    @above_market_price = 500000
   end
 
   def add_room(room)
     @rooms << room
   end
 
-  def add_commas_to_price(price)
-    @price.to_s.reverse.gsub(/(\d{3})(?=\d)/, '\\1,').reverse
+  def reformated_price(price)
+    price.delete('$').to_i
   end
 
   def above_market_price?
@@ -34,7 +34,7 @@ class House
   end
 
   def price_per_square_foot
-    (@price.delete("$").to_i / area.to_f).round(2)
+    (@price / area.to_f).round(2)
   end
 
   def rooms_sorted_by_area
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -5,9 +5,19 @@ class House
     @price = price
     @address = address
     @rooms = []
+    @above_market_price = "$500,000"
   end
 
   def add_room(room)
     @rooms << room
-  end 
+  end
+
+  def add_commas_to_price(price)
+    @price.to_s.reverse.gsub(/(\d{3})(?=\d)/, '\\1,').reverse
+  end
+
+  def above_market_price?
+    @price > @above_market_price
+  end
+
 end
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -2,12 +2,10 @@ class House
   attr_reader :price, :address, :rooms
 
   def initialize(price, address)
-    @attributes = ()
     @price = price
     @address = address
     @rooms = []
     @above_market_price = "$500,000"
-    @details = Hash.new
   end
 
   def add_room(room)
@@ -35,7 +33,22 @@ class House
     key_values.to_h
   end
 
-  
+  def price_per_square_foot
+    (@price.delete("$").to_i / area.to_f).round(2)
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by {|room| room.area}.reverse
+  end
+
+  def rooms_sorted_by_category
+    @rooms.group_by {|room| room.category}
+  end
+
+
+
+
+
 
 
 
--- a/lib/house.rb
+++ b/lib/house.rb
@@ -5,7 +5,6 @@ class House
     @price = reformated_price(price)
     @address = address
     @rooms = []
-    @above_market_price = 500000
   end
 
   def add_room(room)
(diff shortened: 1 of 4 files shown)


lib/house.rb
```diff
@@ -0,0 +1,47 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
```
aziobrow: be careful here-- the interaction pattern asks price to be an integer, so this should be 
`@price = price.to_i`

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+    @address = address
+    @rooms = []
+    @above_market_price = "$500,000"
+  end
+
+  def add_room(room)
+    @rooms << room
```
aziobrow, line 11: 👍

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def add_commas_to_price(price)
+    @price.to_s.reverse.gsub(/(\d{3})(?=\d)/, '\\1,').reverse
```
aziobrow: i don't think there's anything in the interaction pattern that suggests that the price needs to have commas in it... so while I very much appreciate the effort you've shown here, it was not necessary, and the time probably could have been spent better elsewhere.

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def above_market_price?
+    @price > @above_market_price
```
aziobrow: your `@price` and your `@above_market_price` are both strings here... this technically works, but it's more of a happy accident.  Be sure that if you're comparing two values, that you're comparing them as integers, not as strings, since comparing strings can end up with some unexpected behavior.

aziobrow: the idea behind the logic is very good though!

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def rooms_from_category(category)
+    @rooms.find_all {|room| room.category == category}
```
aziobrow, line 23: beautiful. perfect enumerable for the job.

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def area
+    @rooms.sum {|room| room.area}
```
aziobrow, line 27: also beautiful.

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def details(price, address)
```
aziobrow: be careful-- your implementation needs to match the interaction pattern exactly.  The way it asks for the method it be called, it takes no arguments.  There really isn't any need for arguments here anyway, since you have access to `@price` and `@address` as instance variables, so there's no reason to pass them in.

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+    key_values = ["price", price ], ["address", address]
```
aziobrow: This is a little bit unusual, and probably a bit harder to read than the usual way of setting key/value pairs in a hash (but kudos to you for figuring out something new!).  If you did decide to do it this way, in order to match the interaction pattern, you'd have to take the parameters out and this line would look like this instead: 

`key_values = ["price", @price ], ["address", @address]`

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+    key_values.to_h
+  end
+
+  def price_per_square_foot
+    (@price.delete("$").to_i / area.to_f).round(2)
```
aziobrow: very nicely done

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def rooms_sorted_by_area
+    @rooms.sort_by {|room| room.area}.reverse
```
aziobrow, line 40: also SUPER nicely done. very elegant.

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def rooms_sorted_by_category
```
aziobrow: also an excellent solution... except remember to pay attention to those details! This method, according to the interaction pattern, should be called 'rooms_by_category`, so this technically does not pass the spec harness.

test/house_test.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def test_it_exists
+    assert_instance_of House, @house
+  end
+
+  def test_house_has_attributes
+    assert_equal "$400000", @house.price
+    assert_equal "123 sugar lane", @house.address
+    assert_equal ([]), @house.rooms
```
aziobrow: you don't need parentheses around arrays, only around hashes (unless you're going to put them around everything)

test/house_test.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def test_house_can_add_rooms
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    assert_equal [@room_1, @room_2], @house.rooms
+  end
+
+  def test_price_can_be_covereted_using_commas
+    assert_equal "$400,000", @house.add_commas_to_price("$400000")
```
aziobrow: even though you didn't need this method, i'm glad to see a test for it!

test/house_test.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def test_house_is_above_or_below_market_average_price
+    house_1 = House.new("$400000", "123 sugar lane")
+    refute house_1.above_market_price?
+    house_2 =  House.new("$600000", "123 sugar lane")
+    assert house_2.above_market_price?
```
aziobrow, line 36: nice job testing one above and one below

test/house_test.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def test_house_can_list_rooms_from_category
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+    assert_equal ([@room_1, @room_2]), @house.rooms_from_category(:bedroom)
+    assert_equal ([@room_4]), @house.rooms_from_category(:basement)
```
aziobrow: again, not a huge deal, but you don't need parentheses around your expected values here (see, you didn't use them on line 29 of this test, and it was just fine 🙂 )

test/house_test.rb
```diff
@@ -0,0 +1,93 @@
...
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+    @house.add_room(@room_4)
+    assert_equal 1900, @house.area
+  end
+
+  def test_house_can_list_details
+    expected = {"price" => 400000, "address" => "123 sugar lane"}
+    assert_equal expected, @house.details(400000, "123 sugar lane")
```
aziobrow: be sure that the way you assert things in the test match the interaction pattern exactly (i.e., without arguments, in this case)

test/house_test.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def test_rooms_can_be_sorted_by_category
+
+    @house.add_room(@room_4)
+    @house.add_room(@room_1)
+    @house.add_room(@room_2)
+    @house.add_room(@room_3)
+
+    expected = {:bedroom=>[@room_1, @room_2], :living_room=>[@room_3], :basement=> [@room_4]}
+    assert_equal expected, @house.rooms_sorted_by_category
```
aziobrow: make sure to match the name exactly 😬

test/room_test.rb
```diff
@@ -9,16 +9,21 @@ def test_it_exists
...
     assert_equal 130, room1.area
     assert_equal 180, room2.area
   end
+
+  def test_room_starts_unpainted_and_can_be_painted
+    room = Room.new(:bedroom, 10, '13')
+    assert_equal false, room.is_painted?
+    room.paint
+    assert_equal true, room.is_painted?
+  end
```
aziobrow, line 28: nice progression testing here

lib/house.rb
aziobrow: Also, you never actually use this anywhere...

lib/house.rb
```diff
@@ -0,0 +1,47 @@
+class House
+  attr_reader :price, :address, :rooms
+
+  def initialize(price, address)
+    @price = price
+    @address = address
+    @rooms = []
+    @above_market_price = "$500,000"
```
aziobrow: there's no reason for this to be an instance variable. you're never using this outside of the one `above_market_value?` method, so you can just check this value inside of that method.

lib/house.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def add_room(room)
+    @rooms << room
+  end
+
+  def add_commas_to_price(price)
+    @price.to_s.reverse.gsub(/(\d{3})(?=\d)/, '\\1,').reverse
+  end
+
+  def above_market_price?
```
aziobrow: this should be called `above_market_average?` according to the interaction pattern

aziobrow (commented):
Tyler-- overall, very nice work. You did a truly wonderful job of choosing the right tools for the job that you wanted done. 

The methods that you technically missed are price, above_market_average?, details, rooms_by_category... and the only reason you missed all of them is because you didn't pay attention to the interaction pattern as closely as you should have.  See my inline comments for details! 

You're doing well, technically speaking-- you just have to make sure you're paying attention to those little things.. you would have gotten a 4 if not for those careless mistakes. 

